PR #74, opened 2016-12-29
Make GoGet and GoInstall useable from generators

### LINKED ISSUE #73: Export goGet and goInstall from cmd package
In playing around with buffalo I noticed that the goth generator doesn't install the two deps (goth and oauth) when the generator is run.  Unfortunately goGet and goInstall are not exported in the cmd package. 

Instead of generating hard coded exec.Command's for a PR ie:

```go
g.Add(gentronics.NewCommand(exec.Command("go", "get", "-u", "github.com/markbates/goth")))
g.Add(gentronics.NewCommand(exec.Command("go", "get", "-u", "github.com/mrjones/oauth")))
```
 it would be nice/cleaner to use goGet in the goth generator to install the dependencies

intabulas:
goGet/goInstall are un-exported in the cmd package, which makes then unusable for commands on the generate package (ie: goth generator).

This PR adds GoGet and GoInstall to the generate package and updates existing generators to use the new exported version.

This PR also moves the previous goGet/goInstall to root.go (to avoid a package namespace conflict with the soda generator) and changes them to simple call genertate.Go*.

Closes #73


markbates:
Awesome. Thanks!

PR #76, opened 2016-12-29
Install required dependencies when using the Goth generator

### LINKED ISSUE #75: Goth generator does not install required dependencies
When you use the goth generator ex: 
```sh
buffalo g goth twitter facebook linkedin github
```
The two required dependencies ```github.com/markbates/goth``` and ```github.com/mrjones/oauth``` are not installed by default which will cause the server to crash

intabulas:
This PR updates the goth generator (now that #74 is merged) to use go get to install the two required dependencies `github.com/markbates/goth` and `github.com/mrjones/oauth`

Fixes #75


markbates:
Do we need the second go get? or would a go get for `github.com/markbates/goth/...` do the trick?

Also, we should probably add a section to the Dockerfile to generate a new app with goth and make sure it's tests pass. That would've caught this issue a long time ago.

PR #93, opened 2017-01-06
Update build.go

arifemre:
Fixed the build error in Windows operating systems.


markbates:
There looks like there are some code climate issues with this PR.

markbates:
Thanks!

briandowns:
Does `gofmt` leave the external imports intermixed with the stdlib imports?

arifemre:
Not at all!

Issue #59, opened 2016-12-13
`buffalo task` should forward to `grift`


markbates:

markbates:
Closed with #98

PR #98, opened 2017-01-07
making buffalo call grift for the tasks

apaganobeleno:
@markbates this is related to #59. is this what you would expect ?

### DIFF
--- a/buffalo/cmd/task.go
+++ b/buffalo/cmd/task.go
@@ -30,11 +30,7 @@ var taskCommand = &cobra.Command{
 	RunE: func(c *cobra.Command, args []string) error {
 		_, err := exec.LookPath("grift")
 		if err != nil {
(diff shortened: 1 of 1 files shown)


markbates:
@apaganobeleno there are a few code climate issues on this PR. should be pretty quick to fix up though.

apaganobeleno:
@markbates yes, i'll take care of those shortly, thanks for reviewing! should i fix similar other as well ?

markbates:
You are crushing these issues!! Thank you so much!

apaganobeleno:
@markbates thanks for the chance, one question, should we look into writing tests for our cmd package, i don't see that in the repo, but would be something guarantees stability in the future

markbates:
It would definitely be nice. There are smoke tests in the Dockerfile which run some of them.

Issue #16, opened 2016-12-02
Add an "actions" generator


markbates:

apaganobeleno:
@markbates i'm loving buffalo, this is awesome, thanks for taking the lead on this web framework.
Regarding this issue, what do you think should be the output of running 

```bash
buffalo g action myNewAction
```

should there be one action per file inside actions folder ? if so, should  create a new file with
`myNewAction` handler ?

markbates:
Glad you like it!

Here's what I was thinking:

`buffalo g action users show foo bar etc…`

That would create `n` files:

```
actions/users.go
actions/users_test.go
templates/users/show.html
templates/users/foo.html
etc…
```

If `actions/users.go` exists it should append to the file. That file would have the actions:

```go
func UsersFoo(c buffalo.Context) error {
  return c.Render(200, r.HTML("users/foo.html")
}
```

A test asserting the template is rendered would be great as well. 

The template should just be something like:

```html
<h1>Users#Foo</h1>
```

Hopefully that all makes sense. I'm typing on my phone!

Bonus points if you can map the route in `app.go`! :)

apaganobeleno:
@markbates thanks for the explanation on this one its now very clear. I have a question about how to continue with the PR now that i'm contributor, should i send a PR from buffalo's repo or continue sending PR's from my fork?.

Also on another news i've started writing some test cases for this one on `features/actions-generators` i'll keep you posted, but if you have the chance please let me know if i'm going the right direction with the tests.

markbates:
It's up to you whether you would like to create branches here or your fork. Either is fine with me. Just never push directly to master. :)

markbates:
Closed with #103

PR #103, opened 2017-01-09
Adds an Actions generator to the cmd package

apaganobeleno:
@markbates this covers the basic functionality described in #16 it still needs to cover some cases like when the tests signature already exist and the same for the templates, i thought this is a good place to start, i would like to get your feedback.

### DIFF
--- a/buffalo/cmd/generate/action.go
+++ b/buffalo/cmd/generate/action.go
@@ -9,8 +9,8 @@ import (
 	"strings"
 
 	"github.com/aymerick/raymond"
-	"github.com/bep/inflect"
 	"github.com/markbates/gentronics"
+	"github.com/markbates/inflect"
 	"github.com/spf13/cobra"
 )
 
--- a/buffalo/cmd/generate/action.go
+++ b/buffalo/cmd/generate/action.go
@@ -16,8 +16,15 @@ package generate
 
 import (
 	"errors"
+	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
+	"strings"
 
+	"github.com/aymerick/raymond"
+	"github.com/bep/inflect"
+	"github.com/markbates/gentronics"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


buffalo/cmd/generate/action.go
```diff
@@ -0,0 +1,150 @@
...
+		data := gentronics.Data{"under": inflect.Underscore(name)}
+		_, err := os.Stat(filepath.Join("actions", fmt.Sprintf("%v.go", data["under"])))
+		fileExists := err == nil
+
+		g := gentronics.New()
+
+		if !fileExists {
+
+			g.Add(gentronics.NewFile(filepath.Join("actions", fmt.Sprintf("%s.go", data["under"])), rActionFileT))
+			g.Add(gentronics.NewFile(filepath.Join("actions", fmt.Sprintf("%s_test.go", data["under"])), rActionTest))
```
markbates: you need to add the template html file as well.

buffalo/cmd/generate/action.go
```diff
@@ -0,0 +1,141 @@
...
+	return aa
+}
+
+type actionAppender struct {
+	Namespace  string
+	ActionName string
+	FileName   string
+}
+
+func (aa actionAppender) Run(rootPath string, data gentronics.Data) error {
```
markbates: you should add all of this to the gentronics generator.

buffalo/cmd/generate/action.go
```diff
@@ -0,0 +1,141 @@
...
+const (
+	rActionFileT = `package actions
+    import "github.com/gobuffalo/buffalo"`
+
+	rViewT       = `<h1>{{namespace}}#{{action}}</h1>`
+	rActionFuncT = `
+    
+    // {{namespace}}{{action}} default implementation.
+    func {{namespace}}{{action}}(c buffalo.Context) error {
+	    return c.Render(200, r.String("OK"))
```
markbates: this should be `r.HTML("{{namespace}}/{{action}}.html")`

buffalo/cmd/generate/action.go
```diff
@@ -0,0 +1,141 @@
+// Copyright © 2016 NAME HERE <EMAIL ADDRESS>
```
briandowns: Should this be updated before merge?

markbates: yes please. this isn't quite ready to merge yet.

apaganobeleno: @markbates @briandowns what should it say ?

buffalo/cmd/generate/action.go
apaganobeleno: thanks for this one! i missed it.

buffalo/cmd/generate/action.go
markbates: @apaganobeleno whatever is at the top of the other ones? or delete it. don't really care either way. you know what, just delete it. we have a license covering the project. don't need another one there.

buffalo/cmd/generate/action.go
apaganobeleno: I moved this to a `gentronics.Runnable` that will append to existing file, is that ok ? is there a gentronic default runnable that appends to a file ?

buffalo/cmd/generate/action.go
apaganobeleno: @markbates i had to work the actions separately given for the same file will be multiple actions, did i get it correctly ? 🤔

buffalo/cmd/generate/action.go
markbates: no, i guess there isn't. :( but that should be awesome to have! :wink: :wink:

markbates:
@apaganobeleno wrong `inflect` package. :)

```
buffalo/cmd/generate/action.go:12:2: cannot find package "github.com/bep/inflect" in any of:
	/usr/local/go/src/github.com/bep/inflect (from $GOROOT)
	/Users/markbates/Dropbox/development/gocode/src/github.com/bep/inflect (from $GOPATH)
```

should be `github.com/markbates/inflect`

apaganobeleno:
@markbates ah! good catch!

markbates:
@apaganobeleno can you print out the names of the templates that are being generated? keep it uniformed with the rest of the output. You could move those to a generator instead of doing it manually, that will also fix it.

After that, I think we're good to go! :)

apaganobeleno:
@markbates all set on printing the templates. is this how you would like it to look ?
```
Buffalo version 0.7.0

--> actions/cars.go
--> actions/cars_test.go
--> actions/cars.go
--> actions/cars_test.go
--> templates/cars/show.html
--> templates/cars/list.html
--> goimports -w .
```

apaganobeleno:
@markbates now that i gave some thought to what you mentioned i could have built the template to be passed to the generator instead writing my own generator (#needABenderIconHere) i can work on that as a refactoring task if you consider its needed, for now i agree we're good to go.

markbates:
Merged!! Woot! Thank you Antonio!

apaganobeleno:
thanks @markbates i was doing some refactoring and ended up with something, would you like to take a look? i went trough your suggestion of using gentronics and ended up with something simpler, thanks for that, sending another PR for that, it passes the same set of tests.

markbates:
@apaganobeleno yeah, i would love to see something simpler. :) always a fan of simpler.

PR #106, opened 2017-01-10
Using gentronics to generate the templates and the actions

apaganobeleno:
@markbates this should be a bit cleaner and follows your suggestion on using gentronics, please let me know when you have the chance to take a look.

BTW, thanks for those helpers on velvet!


markbates:
that is definitely a lot cleaner! :)

apaganobeleno:
Yeah, sorry for all the back and forwards, holiday down here in 🇨🇴, looking forward to contribute over this week :)

Issue #105, opened 2017-01-10
Need a way to easily add "routes" when generating files.


markbates:
When running something like `buffalo g resource` or `buffalo g actions` it would be nice to able to add those to the routes defined in the `actions/app.go` file.

I propose some sort of function in the `generate` package, probably, that can be called, that properly parses the `actions/app.go` file and allows for the inserting of routes where applicable.

apaganobeleno:
@markbates i would love to give this one a shot, more since this was the bonus part of #16, really enjoying what i'm doing, thanks for the chance.

markbates:
I'm loving the help!! Thanks!

apaganobeleno:
i've been thinking a bit about this one before implementing it, wanted to describe the solution before i implement it to see if we can find any issues to the solution i've been thinking on, i've came with a solution to determine where to place the new route by using `ast` :

https://play.golang.org/p/lCW9eqBr4l

However for me it feels a sensible to the changes we could have on the app.go file, please let me know what do you think, i would love to hear your opinion before starting with the PR.

PR #112, opened 2017-01-12
First Attempt at #105 

apaganobeleno:
hi @markbates i've implemented the first test scenario for the route appender (#105) to the `app.go`, it still assumes some things, but wanted to keep you posted and gather some feedback.


markbates:
I'll take a look at this after I've had some coffee this morning.

markbates:
It would be nice if we could make this a bit more generic. I would like to be able to move this:

https://github.com/gobuffalo/buffalo/blob/master/buffalo/cmd/generate/resource.go#L73

Into app.go as well. Perhaps it takes a string to insert into the file, or function or something. Just spit balling here.

apaganobeleno:
@markbates ok just changed it to receive the expression you want to add inside the block, would be great if you provide me a few other examples, for now i'm assuming everything will go inside the `if app == null` block.

apaganobeleno:
@markbates does this last change helps ?

markbates:
What is the state of this PR? I'd love to merge it in and use it for the next release.

apaganobeleno:
@markbates this one is ready to be used, i also would like to use it for the actions generator.

PR #117, opened 2017-01-12
Default css js files and no logo in assets wo webpack

### LINKED ISSUE #116: Default js and css file when webpack is skipped
I have created a branch in my fork with these 2 default files when webpack is skipped during new project process. I think this is a good idea because in the template they are linked in any case. I added some style to the default page to look a bit similar to the webpack one (no responsive nor font-awesome), just to avoid the big image and the text at the end of the page :). Have a look at it and if you agree I can PR:
https://github.com/fooflare/buffalo/commit/a4c1cf300d937ace42261d2f4e1a65a2a4f479f0

And now, just one question. Why is the logo image generated in ```assets``` and in ```public/assets```? I understand that assets folder is for css and js source files, but I don't get why the logo is saved there too. I ask this because in the case of not using webpack, I think this folder could be omitted.

fooflare:
resolve #116


markbates:
Thanks!

PR #131, opened 2017-01-13
Makes our tests run on Go 1.7 and 1.8

apaganobeleno:
@markbates This PR makes our tests run against Go`1.7` and `1.8` you may notice i set it to be `1.8` instead of `latest` because on the [dockerhub](https://hub.docker.com/_/golang/) latest is 1.7.4 or latest stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ go:
 env:
   matrix:
     - GO_DOCKER_TAG=1.7
-    - GO_DOCKER_TAG=latest
+    - GO_DOCKER_TAG=1.8
 
 script:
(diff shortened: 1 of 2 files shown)


apaganobeleno:
Oh also this should make our build times faster given we're now skipping the travis installation since all we do for our tests runs inside the image we're building.

markbates:
@apaganobeleno so how do i run `docker build .` locally now?

markbates:
@apaganobeleno i made a change to have it run 1.7 locally, and, in theory, get changed in travis. let's see if that works!

apaganobeleno:
Ah! Good one @markbates!

PR #134, opened 2017-01-13
Use glide for the buffalo project and for newly generated apps. 

### LINKED ISSUE #133: Go package dependencies are untracked
Need something like github.com/tools/godep added so go package dependencies are explicit and easy to install with something like godep update

markbates:
closes #133


apaganobeleno:
@markbates should we close this one given we're going to wait for [golang/dep](https://github.com/golang/dep) ?

PR #136, opened 2017-01-14
import buffalo when generating a new action

lumost:
The current action generator does not import buffalo, and the resultant actions will fail to compile.  This PR adds a quick fix and test.


apaganobeleno:
this looks good to me, the actions generator runs `Fmt` task, which check on your system for the `goimports` package if you have it installed it will add the import, i think is better to add it by default and not relay on the `goimports` installation. thanks for this @lumost!

markbates:
Thanks!

PR #160, opened 2017-01-22
Make buffalo sentence syntactically correct with buffalo CLI

dolanor:
Because it is important.

### DIFF
--- a/buffalo/cmd/buffalo.go
+++ b/buffalo/cmd/buffalo.go
@@ -1,0 +1,77 @@
+// Copyright © 2016 NAME HERE <EMAIL ADDRESS>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


buffalo/cmd/buffalo.go
```diff
@@ -0,0 +1,77 @@
+// Copyright © 2016 NAME HERE <EMAIL ADDRESS>
```
philipithomas: This needs to be updated

dolanor: Yes, I copy pasted from another file of the project (I suspect dev.go https://github.com/gobuffalo/buffalo/blob/272bb3a0ab85628caeac9f3ef9e6f0a71f71ef2f/buffalo/cmd/dev.go#L1 ).
My question is (I suck at licensing): should I put my name here, since I wrote part of the code? Or Mark Bates since he is the maintainer?
Another question, the LICENSE at the root indicates it is MIT, but it is a header of apache. So which one should I use?

markbates:
I keep going back and forth on whether to merge this PR.

On the one hand, I think it's funny, and I like that.

On the other hand, it's more code to have in the repo that isn't actually advancing the project. Any one care to weigh in?

dolanor:
I can weigh in. The problem with that, is that it might confuse new comers to the CLI. So actually, I think this PR is a bad idea.
It's just that you inspired me to do it, and I promised to send a PR :stuck_out_tongue:. I won't take it personally if you don't merge it.
And I thought I could improve it and use a slice of `buffalo` instead of a loop with some exception to deal with. It should be more readable and simple.

Though, my question about the licensing header still holds as I don't know the answer, so if anyone can enlighten me?

markbates:
@dolanor well, if you don't think it should be merged than that's good enough for me. :)

As for the licensing, Cobra automatically adds that in at the top when a new command is generated. It should probably all be removed since the project is MIT, we don't need the license at the top of every command.

I'll open a ticket for it.

PR #162, opened 2017-01-23
Updated Unknown Error message

bscott:
Updated the "Unkown" error message, redirecting users to review the Troubleshooting guide for tips.  The Troubleshooting Page will have tips on Node/NPM install. 

PR will be issues for the following page.
http://gobuffalo.io/docs/troubleshooting

Thoughts?


markbates:
Yeah, point them at the troubleshooting docs. That seems like a great place to start.

PR #163, opened 2017-01-23
Added jetbrains IDE workspace directory for such IDE's as Gogland

bscott:
Added workspace directory for the JetBrains line of IDE


markbates:
You should add this to the generated .gitignore as well.

bscott:
@markbates Just added the updated commit

PR #172, opened 2017-01-24
Fix alternative layout (cont'd)

XelaRellum:
Sorry, seems like the last pull request broke some stuff in some cases and was missing a testcase. Fixed this.


markbates:
Thanks! I saw your original PR and fixed it with #171. So this is all good. Thanks.

Issue #36, opened 2016-12-11
generate a default .travis.yml file for new projects


markbates:

briandowns:
Should there be a flag similar to `--db-type` for CI provider? e.g. `--ci-provider` and support Travis and Circle out of the box? Maybe add others depending on demand?

markbates:
That seems like a good idea.

briandowns:
I have a branch started to complete the default travis file but if the additional flag is an acceptable approach, I'd like to include that as well.  Also, was hoping for feedback on implementation: cobra command to generate the `buffalo g` flags mentioned above.

markbates:
If there's only CI we don't need a flag, for now. When we want different CI files, then we can add a flag. 

There are plenty of examples of how to use flags in the cmd directory.

briandowns:
I can add that flag functionality for different CI providers in this branch if you want.  Your call. :)

markbates:
I'd love it if you feel comfortable enough generating files for other CIs!

markbates:
Those two would be great! Wercker would be awesome as well,  if you can get to it.

apaganobeleno:
@briandowns are you working on this one ? i have some time to give it a shot, but if you're working on it prefer you to complete it :)

cdrx:
Is there an example of a working circle.yml or travis-ci.yml file?

briandowns:
I'll try to get this pushed up today.

stanislas-m:
Gitlab CI support could be cool too. :)

markbates:
A wercker set up would be great too. These should be separate tickets though.

apaganobeleno:
@markbates i started working on this the `.travis.yml` however when running the tests for the first time its panic'ing with:

```
--- FAIL: Test_HomeHandler (0.00s)
panic: could not locate box "../public/assets" [recovered]
	panic: could not locate box "../public/assets"
```

Should we remove the `.gitignore` from that folder or handle rice diferently when running tests ?
Also for the travis configuration, i'm assuming we should run migrations/setup database and run the tests with `buffalo test`, am i right ?

apaganobeleno:
@markbates nevermind, i ended up creating the folder on the `.travis.yml`

PR #184, opened 2017-01-26
[feature] adds .travis.yml when generating a new app

apaganobeleno:
Adds .travis.yml to the newly generated project as part of #36 

This also allows to avoid ci file being generated by using the `--ci-provider=none` flag

cc @markbates


markbates:
I think I would prefer the default to be no CI and let people choose. Thoughts?

apaganobeleno:
@markbates i'm ok with that, lets change it.

buffalo/cmd/app_generators.go
```diff
@@ -361,3 +366,18 @@ ratings:
     - "**.go"
 
 `
+
+const nTravis = `language: go
+env:
+- GO_ENV=test
+
+before_script:
```
markbates: this raises an interesting question about databases here. this is specific to Postgres, but what if they generate with mysql or sqlite?

apaganobeleno: Oh! Good catch @markbates

apaganobeleno: i'll work on this soon and keep you posted.

philipithomas:
Travis has a linter for their .travis.yml files - I think that it should be incorporated somewhere into testing for this

apaganobeleno:
@philipithomas validating the .travis.yml sounds like a great idea, however would imply installing ruby inside the docker container as well as the travis gem, @markbates what do you think? should we add that validation ?

markbates:
If we validate the file we're embedding here, I think that's good enough for now. Unless the have an API we can ping or something.

apaganobeleno:
@markbates great, i don't know of an API for that, just to clarify, when you say validate, you mean we check in our tests the file being created and its content, right ? something else we should check on that validation ?

apaganobeleno:
@markbates @philipithomas seems there is one :) https://docs.travis-ci.com/api#linting, i'll work on this one over the week.

apaganobeleno:
As an update while using the yml linter i found an [issue](https://github.com/travis-ci/travis-yaml/issues/110) with the validator (at least the API) when using Curl.

markbates:
@apaganobeleno is this ready to merge, or are you still working on it?

apaganobeleno:
@markbates this one is ready to merge, was looking for a way to have some tests that could cover running the new command and passing the --ci-provider travis flag, but the GOPATH check on the command is making it hard, anyways we need to add some tests there so, i think we can cover that lately.

PR #185, opened 2017-01-27
fixes #183

### LINKED ISSUE #183: Silent failure of unknown db type
If you run the following.

`buffalo new coke --db-type SOMETHING-BAD! > result.out 2>&1`

It seems like the db setup fails silently due to the unknown type being passed. I don't see a warning in the output file.

The issue seems to be around here https://github.com/gobuffalo/buffalo/blob/master/buffalo/cmd/soda_generators.go#L34.

`generate.GenerateConfig("./database.yml", data)` returns an error but is being ignored and instead on line 35 nil is being returned silencing the error.

Possible solution is to return the error from GenerateConfig.

```go
g.Add(&gentronics.Func{
		Should: should,
		Runner: func(rootPath string, data gentronics.Data) error {
			data["dialect"] = data["dbType"]
			return generate.GenerateConfig("./database.yml", data)
		},
	})
```

This change results in the following output but does not fail the new app from generating. However, it does give the user some information to know what went wrong.

`Error: Could not initialize something-bad!!`

However, a better solution is to return the error but also validate the arguments being supplied by the user and bail out if they are invalid with a helpful message.

I can come up with a pull request for this if this sounds good.

amedeiros:
Fixes the silent failing of a unknown db-type. The following is the result.

```
buffalo new coke --db-type SOMETHING-BAD!

Buffalo version 0.7.1

Error: Unknown db-type SOMETHING-BAD! expecting one of postgres, mysql or sqlite3
Usage:
  buffalo new [name] [flags]

Flags:
      --db-type string   specify the type of database you want to use [postgres, mysql, sqlite3] (default "postgres")
  -f, --force            delete and remake if the app already exists
      --skip-pop         skips adding pop/soda to your app
      --skip-webpack     skips adding Webpack to your app
  -v, --verbose          verbosely print out the go get/install commands
```

### DIFF
--- a/buffalo/cmd/root.go
+++ b/buffalo/cmd/root.go
@@ -44,7 +44,6 @@ var RootCmd = &cobra.Command{
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	if err := RootCmd.Execute(); err != nil {
-		fmt.Println(err)
 		os.Exit(-1)
 	}
 }
--- a/buffalo/cmd/new.go
+++ b/buffalo/cmd/new.go
@@ -78,12 +78,7 @@ var newCmd = &cobra.Command{
 }
 
 func validDbType() bool {
-	switch dbType {
-	case "postgres", "mysql", "sqlite3":
-		return true
-	default:
-		return false
-	}
+	return dbType == "postgres" || dbType == "mysql" || dbType == "sqlite3"
 }
 
 func validateInGoPath(name string) error {
--- a/buffalo/cmd/new.go
+++ b/buffalo/cmd/new.go
@@ -44,6 +44,10 @@ var newCmd = &cobra.Command{
 	Use:   "new [name]",
 	Short: "Creates a new Buffalo application",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if !validDbType() {
+			return fmt.Errorf("Unknown db-type %s expecting one of postgres, mysql or sqlite3", dbType)
+		}
+
 		if len(args) == 0 {
 			return errors.New("you must enter a name for your new application")
 		}
@@ -71,6 +75,15 @@ var newCmd = &cobra.Command{
 
 		return genNewFiles(name, rootPath)
 	},
+}
+
+func validDbType() bool {
+	switch dbType {
+	case "postgres", "mysql", "sqlite3":
+		return true
+	default:
+		return false
+	}
 }
 
 func validateInGoPath(name string) error {
--- a/buffalo/cmd/soda_generators.go
+++ b/buffalo/cmd/soda_generators.go
@@ -31,8 +31,7 @@ func newSodaGenerator() *gentronics.Generator {
 		Should: should,
 		Runner: func(rootPath string, data gentronics.Data) error {
 			data["dialect"] = data["dbType"]
-			generate.GenerateConfig("./database.yml", data)
-			return nil
+			return generate.GenerateConfig("./database.yml", data)
 		},
 	})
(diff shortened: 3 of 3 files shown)


buffalo/cmd/root.go
```diff
@@ -44,7 +44,6 @@ var RootCmd = &cobra.Command{
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	if err := RootCmd.Execute(); err != nil {
-		fmt.Println(err)
```
markbates: why did you remove the printing of the error here? or was that a mistake?

amedeiros: I removed it because it was causing the error to be printed twice.

amedeiros: Instead of the following

```
buffalo new coke --db-type SOMETHING-BAD!
Buffalo version 0.7.1

Error: Unknown db-type SOMETHING-BAD! expecting one of postgres, mysql or sqlite3
Usage:
  buffalo new [name] [flags]

Flags:
      --db-type string   specify the type of database you want to use [postgres, mysql, sqlite3] (default "postgres")
  -f, --force            delete and remake if the app already exists
      --skip-pop         skips adding pop/soda to your app
      --skip-webpack     skips adding Webpack to your app
  -v, --verbose          verbosely print out the go get/install commands```

You would get the error out put again at the end. Shown below.

```
buffalo new coke --db-type SOMETHING-BAD!
Buffalo version 0.7.1

Error: Unknown db-type SOMETHING-BAD! expecting one of postgres, mysql or sqlite3
Usage:
  buffalo new [name] [flags]

Flags:
      --db-type string   specify the type of database you want to use [postgres, mysql, sqlite3] (default "postgres")
  -f, --force            delete and remake if the app already exists
      --skip-pop         skips adding pop/soda to your app
      --skip-webpack     skips adding Webpack to your app
  -v, --verbose          verbosely print out the go get/install commands

Unknown db-type SOMETHING-BAD! expecting one of postgres, mysql or sqlite3```

markbates:
looks good. thanks!

amedeiros:
@markbates Thanks for the merge. I was just messing with error message situation. If setting the root command to silence errors we can add that println back and it will instead print the error at the end. I guess its a trade off on where the message gets printed really.

```go
// RootCmd is the hook for all of the other commands in the buffalo binary.
var RootCmd = &cobra.Command{
	SilenceErrors: true,
	Use:   "buffalo",
	Short: "Helps you build your Buffalo applications that much easier!",
	PersistentPreRun: func(cmd *cobra.Command, args []string) {
		fmt.Printf("Buffalo version %s\n\n", Version)
	},
}

// Execute adds all child commands to the root command sets flags appropriately.
// This is called by main.main(). It only needs to happen once to the rootCmd.
func Execute() {
	if err := RootCmd.Execute(); err != nil {
		fmt.Println(fmt.Sprintf("Error: %s", err))
		os.Exit(-1)
	}
}
```

That will result in the following output instead.

```
buffalo new coke --db-type SOMETHING-BAD!
Buffalo version 0.7.1

Usage:
  buffalo new [name] [flags]

Flags:
      --db-type string   specify the type of database you want to use [postgres, mysql, sqlite3] (default "postgres")
  -f, --force            delete and remake if the app already exists
      --skip-pop         skips adding pop/soda to your app
      --skip-webpack     skips adding Webpack to your app
  -v, --verbose          verbosely print out the go get/install commands

Error: Unknown db-type SOMETHING-BAD! expecting one of postgres, mysql or sqlite3
```

PR #189, opened 2017-01-28
Add Golint and fix all issues

philipithomas:
Go Report Card reported 26 linting issues with Buffalo:

https://goreportcard.com/report/github.com/gobuffalo/buffalo

I went through and corrected all of them, plus added linting to the Travis test. 

---

@markbates As discussed, I just made standard golint pass. You may want to keep grifts imports as a top-level rather than namespaced. (I personally found this confusing because of how many different type of "Context" types get passed around)

So - if you want me to do some modifications to allow for dot imports and revert those files, let me know.

### DIFF
--- a/examples/hello-world/grifts/routes.go
+++ b/examples/hello-world/grifts/routes.go
@@ -4,11 +4,11 @@ import (
 	"os"
 
 	"github.com/gobuffalo/buffalo/examples/hello-world/actions"
-	. "github.com/markbates/grift/grift"
+	mbgrift "github.com/markbates/grift/grift"
 	"github.com/olekukonko/tablewriter"
 )
 
-var _ = Add("routes", func(c *Context) error {
+var _ = mbgrift.Add("routes", func(c *mbgrift.Context) error {
 	a := actions.App()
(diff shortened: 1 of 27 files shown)


examples/hello-world/grifts/routes.go
```diff
@@ -4,11 +4,11 @@ import (
 	"os"
 
 	"github.com/gobuffalo/buffalo/examples/hello-world/actions"
-	. "github.com/markbates/grift/grift"
+	mbgrift "github.com/markbates/grift/grift"
 	"github.com/olekukonko/tablewriter"
```
markbates: Why are we using a special name here?

markbates:
Other than the one question I have, this looks good.

examples/hello-world/grifts/routes.go
philipithomas: Whoops, just pushed a commit to fix it. 

when I first started this PR . . . I misread something and thought there was a namespace collision. But there's not.

philipithomas:
@markbates Corrected

markbates:
Thanks!

Issue #198, opened 2017-01-30
Resource generator shouldn't generate pointers


markbates:
Currently the resource generator generates the interface on a pointer type, this is unnecessary, it should be a value type:

```go
type TodosResource struct {
	buffalo.Resource
}

// List default implementation.
func (v TodosResource) List(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#List"))
}

// Show default implementation.
func (v TodosResource) Show(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#Show"))
}

// New default implementation.
func (v TodosResource) New(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#New"))
}

// Create default implementation.
func (v TodosResource) Create(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#Create"))
}

// Edit default implementation.
func (v TodosResource) Edit(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#Edit"))
}

// Update default implementation.
func (v TodosResource) Update(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#Update"))
}

// Destroy default implementation.
func (v TodosResource) Destroy(c buffalo.Context) error {
	return c.Render(200, r.String("Todos#Destroy"))
}
```

PR #200, opened 2017-01-31
[#198] moving resource generator to avoid pointers

apaganobeleno:
Avoids using pointers for generated resource functions.
cc @markbates


markbates:
We should change what gets generated in the app.go file.

Issue #193, opened 2017-01-30
Add file:line info to Context#Get deprecation warnings


markbates:

apaganobeleno:
@markbates is this all-set with #193?

Issue #153, opened 2017-01-21
buffalo.Context should implement the context.Context interface


markbates:
Since this would involve a couple of changes that will eventually change the `buffalo.Context` API, we'll need to make sure to deprecate the functions that get changed, particularly `Get` and `Set`, and remove them in a later version of Buffalo.

PR #202, opened 2017-01-31
[#193] adding line number to the Context#Get warning

apaganobeleno:
cc @markbates

### DIFF
--- a/default_context.go
+++ b/default_context.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


bscott:
@apaganobeleno I'm just catching up, why did we deprecate `context#get`

apaganobeleno:
@bscott this is related with https://github.com/gobuffalo/buffalo/issues/153

apaganobeleno:
@markbates something else this one needs ? better wording p.e. ?

markbates:
@bscott i deprecated `Get` because I implemented the `context.Context` interface, and didn't see the need for both `Get` and `Value`. I prefer the name `Get`, but it seemed to silly to have both.

Issue #210, opened 2017-02-01
Add support for "named routes"


markbates:
Since we're using http://www.gorillatoolkit.org/pkg/mux under the covers we could easily create named routes for everything. When we create a new route we return the RouteInfo for it so we could hang a Name function on that calls the underlying mux name function.

We would also need to expose those in the context.

Related to #207 

Example:
```go
app.GET("/users") // -> users_path
app.POST("/users") // -> users_path
app.GET("/users/new") // -> new_users_path
app.GET("/users/{user_id}") // -> user_path 
app.GET("/users/{user_id}/children") // -> user_children_path 
app.GET("/users/{user_id}/children/{child_id}") // -> user_child_path 
// etc ...

app.GET("/people").Name("peeps") // -> peeps_path
```

apaganobeleno:
Resources should have this too, right @markbates ?

markbates:
Of course!

bscott:
@apaganobeleno Are you doing this one?, can I assign it to you?

apaganobeleno:
oh sir i would love to, yes please, unless @markbates is doing it please assign to me :)

markbates:
@apaganobeleno it's all yours!

apaganobeleno:
Hey @markbates i'm working on this over the weekend, i wanted to summarize the behavior here now that plush is in place and looking forward to integrate this into v0.8.0.

This feature is willing to provide users with the chance to do something like:

```go
app.GET("/users")
app.GET("/people").Name("peeps") 
app.GET("/people/{user_id}").Name("peeps") 
```
What will provide 3 functions that the user could use like:

```erb
<a href="<%= usersPath() %>">Users</a>
<a href="<%= peepsPath() %>">Peeps</a>
<a href="<%= peepPath(randomPeepID) %>">Random Peep</a>
```
As can be noticed route name can be defined explicitly (peepPath) or it will be generated by buffalo automatically.

Am i right ?

markbates:
That's mostly right. The only thing I would say is that the path functions should take a map.

```go
<a href="<%= usersPath() %>">Users</a>
<a href="<%= peepsPath() %>">Peeps</a>
<a href="<%= peepPath({user_id: randomPeepID}) %>">Random Peep</a>
```

A little secret, that most people don't know is that if `map[string]interface{}` is the last, or second to last argument in a Plush helper function, then it is optional. :)

```go
func peepPath(opts map[string]interface{}) {
// opts is optional
}
```

[...]

PR #216, opened 2017-02-03
Implementing route names generator and setter

apaganobeleno:
@markbates this implements RouteInfo Name and name builder based on the examples you provided on #210.

I still don't have very clear what to expose in the context, would you like to expose the current route or just add the route name on the context data, i also wonder if the `current_route` thats being set in the context data could be sufficient now that `RouteInfo` has the `RouteName` method which returns the route name.

Thoughts ? cc @bscott


apaganobeleno:
This one is ready to be merged.
cc @markbates.

apaganobeleno:
@markbates i take that down, still need to work on the RouteName method.

markbates:
lol. It didn't seem quite ready just yet. :)

We also need a way of getting access to the them in the views, and potentially in the controller.

apaganobeleno:
Ok, now its taking shape, for now we are allowing developers to name routes by doing:
```go
app.GET("/", OverviewHandler).RouteName("overview")
```
and by setting the current route into `data["current_route"]` on the view we are now able to do:
```go
```

apaganobeleno:
@markbates congratz on Patriots victory, amazing game! Just added a CurrentRoute method on the context, do you think this solves the actions access to the current_route ?.

app.go
```diff
@@ -85,3 +88,11 @@ func Automatic(opts Options) *App {
 
 	return a
 }
+
+func (a *App) addDefaultRouteNames() {
```
markbates: why would we need this? shouldn't all routes get names when they're first added?

context.go
```diff
@@ -33,6 +33,7 @@ type Context interface {
 	Redirect(int, string, ...interface{}) error
 	Data() map[string]interface{}
 	Flash() *Flash
+	CurrentRoute() RouteInfo
```
markbates: Let's try and not add more things to the context unless we really need it. The `current_route` is already in the context data, if people need it in a handler they can get it that way. If it becomes a problem later we can extend the context, but i'm already nervous as to how big it already is.

default_context.go
```diff
@@ -209,3 +210,8 @@ var defaultUpgrader = websocket.Upgrader{
 	WriteBufferSize: 1024,
 	CheckOrigin:     func(r *http.Request) bool { return true },
 }
+
+// CurrentRoute returns the current route information.
+func (d *DefaultContext) CurrentRoute() RouteInfo {
```
markbates: See my other comment about removing this from the context

app.go
apaganobeleno: @markbates  Oh, yes i forget to talk about this, we need do it after because mux doesn't allow a route to name be changed after it was set ([look here](https://github.com/gorilla/mux/blob/34dda716af12ba79ed40ce1f66b102ff75dc3411/route.go#L115)).

context.go
apaganobeleno: ok, i'll remove this one, thanks for the feedback!

apaganobeleno:
@markbates removed CurrenRoute from context, does my explanation for `addDefaultRouteNames` make sense? would you like us to stick with mux names or handle names in buffalo's RouteInfo ?

apaganobeleno:
@markbates do we need something else here ? should we move to store route names in RouteInfo ?

app.go
markbates: I still don't quite understand what this means, sorry. Also, the fact it happens on every request isn't great either.

markbates:
We're still missing a way to actually use these named routes.

I'm thinking a map of some kind, `map[string]routefunc`:

```handlebars
<a href="{{ urls.user_path user_id=current_user.ID }}">{{current_user.Name}}</a>
```

something like that.

app.go
apaganobeleno: @markbates I'll try to explain myself a bit better, sorry for the confusion, but first let me mention some constraints i've understood we have on this feature:

1. We want to use gorilla's mux named features
2. We don't want to add more code to the actions.go file

To start, right now we have the following code to add a route into our application:

```go
app.GET("/", SomeHandler)
```

What i was mentioning before is that it seems that we cannot add the default route name on the mux Route thats inside the RouteInfo (if none passed) by the time GET/POST(...) method is called because by doing so we would be disallowing the RouteName method to add the route and invalidating the entire route (because of [this](https://github.com/gorilla/mux/blob/34dda716af12ba79ed40ce1f66b102ff75dc3411/route.go#L115)).

I agree that this happening on every request is not good but i still looking for a better place to do it without either adding more code into the generated actions.go or storing the route name outside the *mux.Route.

i hope this makes more sense!

markbates:
Ok, I think we need to rethink this feature a bit. Doesn't seem like using the mux named routes is going to work out the way we want.

Let me go back to the drawing board and think about this a bit.

apaganobeleno:
thanks @markbates, i've been bitten by the flu over the weekend, but looking forward to work on this one, should we close this PR ?

markbates:
Hope you're feeling better. Don't close it. I started playing around with it over the weekend, but it lead to a limitation in the tools I'm using to build velvet (particularly the ast, lexer, and parser from the Raymond library I'm using). That lead me to start rebuilding the tools from scratch to make velvet more powerful and easier to work with so I can get back to the view side of this issue. :)

apaganobeleno:
Looking forward for Plush integration to continue with this one.

markbates:
closing this in favor of #332

PR #218, opened 2017-02-03
Buffalo new "welcome" output closes #164

### LINKED ISSUE #164: Buffalo new "welcome" output
After running `buffalo news there should be some nice "next steps" print out. For example:

* path to the new app
* updating the database.yml
* set up "postgres|mysql|sqlite|
* buffalo dev
Etc...

Those are just some ideas, but it would be nice to pint people in the right direction.

markbates:


apaganobeleno:
this one sounds like 1.7.2 🚀

PR #234, opened 2017-02-06
content types need to be ranged over in case of ones with a ';' in them

markbates:


markbates:
Ah, yes, good point. Didn't even cross my mind. Good catch.

default_context.go
```diff
@@ -161,11 +161,9 @@ func (d *DefaultContext) Render(status int, rr render.Renderer) error {
 // binder is "http://www.gorillatoolkit.org/pkg/schema".
 func (d *DefaultContext) Bind(value interface{}) error {
 	ct := strings.ToLower(d.Request().Header.Get("Content-Type"))
-	cts := strings.Split(ct, ";")
-	for _, c := range cts {
-		if b, ok := binders[strings.TrimSpace(c)]; ok {
-			return b(d.Request(), value)
-		}
+	cts := strings.Split(ct, ";")[0]
```
philipithomas: Did you test this for empty string "ct"? I'm afraid it will panic

markbates: Splitting on an empty string returns a slice that has 1 empty string element in it:

https://play.golang.org/p/_y55t5_kC9

Still, I'll wrap it anyway, just to prevent future peeps from questioning at it.

philipithomas:
@markbates I opened a PR into this branch to check the more specific scenario that I was thinking of. It looks like it did not error, so I learned something today :-)

https://github.com/gobuffalo/buffalo/pull/238

default_context.go
vasukolli: if content type contains charset like "application/json; charset=utf-8", then charset will not be considered.
if I am wrong please correct me.

PR #235, opened 2017-02-07
removed the no longer existing docs for LogDir and added some for

markbates:
MethodOverride in Options


u007:
hi, how do i tell buffalo to generate log on staging environment in maybe log/staging.log?

markbates:
See https://github.com/gobuffalo/buffalo/issues/519

PR #239, opened 2017-02-07
get latest dependencies on `new` command

philipithomas:
I ran into an issue where Buffalo updated, and that update relied on a change in another repo that had not been updated. 

Pros: 
* Makes buffalo easier to use
* Keeps dependencies up to date until we use a dependency manager

Minuses:
* `new` command is slightly slower
* `new` may not work without internet

Regarding the need for internet - I think that `npm install` may require internet, anyways?


markbates:
This was originally like is, but it fails if you've got uncommitted changes locally and you've built the binary yourself. 

Scenario:
You're working on some feature of buffalo in a local branch. You build the binary to test that feature. You run buffalo new. The -u flag yells and stops.

philipithomas:
Interesting. Most users won't be modifying dependencies, and if they don't have knowledge of the source - resolving a dependency issue is not trivial (because there's no "update dependencies"  command or list of deps). 

I propose two possible solutions:

* Keep the `-u` on `go gets` for `buffalo new`, but add a command line flag to `buffalo new` that skips the `go get`s, `npm install`s, etc (e.g. maybe a `-q` or `--quick`)
* When calling `buffalo new`, recompile from source, and if there is a dependency error - update all dependencies

markbates:
A flag to turn off 'go get' seems like a good option. There is already a flag to skip webpack install.

markbates:
@philipithomas can we add a `--no-update` flag to `buffalo new`? That will let people turn off updating during `go get`. I would love to get this merged in for `0.7.3`.

philipithomas:
Yes. I'll update this PR hopefully by Monday

jrnt30:
New user to Buffalo.  Thanks for the work and really just looking for context, not to start a holy war if this has been brought up already. 

> Keeps dependencies up to date until we use a dependency manager

Is this on the roadmap anytime soon?  
Are there reasons for not just "choosing one" for now?

stanislas-m:
@jrnt30 glide was about to be used, but it was decided to "wait for https://github.com/golang/dep to be released so we don't have to change again". See #133.

PR #240, opened 2017-02-11
Allow go get build flags (will be useful with soda generators).

stanislas-m:
Related to https://github.com/markbates/pop/issues/27.

Will allow buffalo to use the only needed database driver dependencies.

### DIFF
--- a/buffalo/cmd/soda_generators.go
+++ b/buffalo/cmd/soda_generators.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"github.com/markbates/gentronics"
 	"github.com/markbates/pop/soda/cmd/generate"
+	"github.com/pkg/errors"
 )
 
 func newSodaGenerator() *gentronics.Generator {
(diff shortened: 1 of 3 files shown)


markbates:
I like this, but it's not quite there yet. First, I don't see how the actual build tags get to those functions. They accept them, but nothing ever calls.

Also, it would be great if `buffalo build` can take build flags and pass them to `go build` command.

stanislas-m:
@markbates 

> I like this, but it's not quite there yet. First, I don't see how the actual build tags get to those functions. They accept them, but nothing ever calls.

Here it is.

> Also, it would be great if buffalo build can take build flags and pass them to go build command.

Sure. I'll take a look.

buffalo/cmd/soda_generators.go
```diff
@@ -19,11 +20,25 @@ func newSodaGenerator() *gentronics.Generator {
...
 	g.Add(f)
 
-	c := gentronics.NewCommand(goGet("github.com/markbates/pop/..."))
-	c.Should = should
-	g.Add(c)
+	g.Add(&gentronics.Func{
+		Should: should,
+		Runner: func(rootPath string, data gentronics.Data) error {
+			var driver string
+			if data["dbType"] == "sqlite3" {
```
markbates: just a thought, but if we drop the `_driver` from the build tag then we don't need this if statement.

also, it might be possible that people have two different DBs, say SQLite in development or test and Postgres in production. That's more of a thing we need to worry about during `buffalo build`. For that we can look at the `database.yml` file during build time and grab all of the dialects being used there to pass as flags.

stanislas-m: > just a thought, but if we drop the _driver from the build tag then we don't need this if statement.
Sure. I've dropped the suffix on the pop side, so I've just to adjust the buffalo side now.

> also, it might be possible that people have two different DBs, say SQLite in development or test and Postgres in production. That's more of a thing we need to worry about during buffalo build. For that we can look at the database.yml file during build time and grab all of the dialects being used there to pass as flags.

Hum, I already thought about reading the `database.yml` for the build part, I'm glad you thought about it too. In this case, maybe we can provide the whole drivers set for the development part, and take care of required drivers only on build time? (and in this case, I can add a special `all` flag, to provide all drivers support, without having to enumerate them all)

stanislas-m:
This one will need further investigation, I'll reopen another better pull request when it's ready.

Issue #187, opened 2017-01-27
Proposal: Use `yarn` instead of `npm install`


philipithomas:
Yarn is a drop-in replacement from Facebook for `npm install`. The advantage is that is deterministic in its build, and it pulls from a local cache (similar to how the [glide](http://glide.sh) package manager works). 

The advantage is that it is significantly faster than 

**Proposal**:

* When generating new files, include a `yarn.lock` file. (On TravisCI, [this triggers the use of yarn](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn))
* Check if `yarn.lock` file exists locally. If it does, use `yarn` to install modules.  
* If no `yarn.lock` file is present, check if `yarn` is installed locally. If it is, use `yarn`
* Fallback to `npm install` if yarn is unavailable

markbates:
I would like to hear about the disadvantages as well. Everyone I've talked to who has used yarn has had to go back to NPM for various reasons. Currently I don't know anyone who uses it in production or on bigger projects. I definitely need to be sold more on it.

philipithomas:
We've been using Yarn for awhile without issue. I'm curious to hear about others' experiences. 

What would you think of having Buffalo check for the presence of a `yarn.lock` file, and if and only if it exists - using `yarn` instead of `npm install`? I would personally use that on projects.

markbates:
That doesn't help much when generating a new project. I'd be open to a second asset generator and let people pick if they want npm or yarn or neither when building a new app. The default should be npm since it's still the "standard" for whatever that term means in JS land. :)

philipithomas:
Ok. I should have some time this weekend for a test implementation. Could probably break it into two PRs - the first for detecting a yarn.lock during the build process, and the second for modifying the `buffalo create` tool.

markbates:
Ah. The build part that's why you would look for the lock file. Gotcha. Didn't think of that.

markbates:
Buffalo dev would need to be updated as well.

gillchristian:
I was going to open an issue for this as well. If you need any help @philipithomas let me know :smile: 

_Btw_, I've been using `yarn` in production too with no issues.

fooflare:
What about adding a --with-yarn option and keep npm as default? The file yarn.lock would be generated during the first build, so you can check if it exists in the next ones. Is this going forward? If any help is needed, I am open to work on it.

markbates:
Closed via #256

PR #256, opened 2017-02-21
Add yarn as alternative --with-yarn

fooflare:
As I mentioned in #187 I made some changes to have yarn as option with --with-yarn flag. It works following the specifications below:

1. With new:
  1.1. If flag is specified:
      * Checks if yarn is installed and if not, installs it and changes the default command from npm to yarn (adding equivalent arguments)
      * Inits the yarn project, downloads and installs webpack dependencies

     1.2. If not, npm is executed

2. With generate webpack, I added the same flag as alternative:
  2.1. Checks if yarn.lock already exists and if so, yarn is established as default
  2.2. If yarn.lock doesn't exist:
     * If flag was specified, yarn is selected (same process than with new command)
     * If not, npm is executed

### DIFF
--- a/buffalo/cmd/generate/webpack.go
+++ b/buffalo/cmd/generate/webpack.go
@@ -2,6 +2,7 @@ package generate
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


markbates:
This looks good, but to be feature complete it needs to support the following commands:

- [ ] buffalo new
- [ ] buffalo dev
- [ ] buffalo build

Those commands are all touch the "asset pipeline".

I also think we should rename `--skip-webpack` to `--skip-assets`.

fooflare:
One thing I haven't clear is about dev and build. I checked both but is webpack directly executed, not npm. Or am I wrong?

In reference to the renaming, I am not sure if --skip-assets or maybe --skip-npm. In this PR if webpack is skipped, yarn is skipped too, even if the flag is specified (no dependencies to be generated). And assets are generated in public in any case (application.js/.css and logo), so I am not sure if it would be totally clear.

markbates:
Ok, I tested this and validated that all three commands work with a `yarn` application. Fantastic. Thanks for this.

fooflare:
Don't thank me, I did very few. Thanks to you for all your efforts in the project.

markbates:
It's a group effort! :)

philipithomas:
👍

Issue #262, opened 2017-02-26
v0.7.3 webpack2.0 release breaks default configuration


charlesmims:
With a fresh install of v0.7.3 and go v1.8, using postres, I get the following error when trying to start `buffalo dev` on a new project:

```
cmims@MBP:~/go/src/github.com/charlesmims/buffalo-blog$ buffalo dev
Buffalo version 0.7.3

buffalo: 2017/02/26 13:54:39 === Rebuild on: :start: ===
buffalo: 2017/02/26 13:54:39 === Running: go build -v -i -o tmp/buffalo-blog-build (PID: 14209) ===
/Users/cmims/go/src/github.com/charlesmims/buffalo-blog/node_modules/extract-text-webpack-plugin/index.js:176
		throw new Error("Breaking change: extract now only takes a single argument. Either an options " +
		^

Error: Breaking change: extract now only takes a single argument. Either an options object *or* the loader(s).
Example: if your old code looked like this:
    ExtractTextPlugin.extract('style-loader', 'css-loader')

You would change it to:
    ExtractTextPlugin.extract({ fallback: 'style-loader', use: 'css-loader' })

The available options are:
    use: string | object | loader[]
    fallback: string | object | loader[]
    publicPath: string

    at Function.ExtractTextPlugin.extract (/Users/cmims/go/src/github.com/charlesmims/buffalo-blog/node_modules/extract-text-webpack-plugin/index.js:176:9)
    at Object.<anonymous> (/Users/cmims/go/src/github.com/charlesmims/buffalo-blog/webpack.config.js:38:33)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at module.exports (/Users/cmims/go/src/github.com/charlesmims/buffalo-blog/node_modules/webpack/bin/convert-argv.js:80:13)
    at Object.<anonymous> (/Users/cmims/go/src/github.com/charlesmims/buffalo-blog/node_modules/webpack/bin/webpack.js:39:40)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
[...]

Issue #195, opened 2017-01-30
Buffalo needs to be updated to support Webpack 2.2.x


markbates:

markbates:
```
/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/webpack/lib/Chunk.js:59
		throw new Error("Chunk.entry was removed. Use hasRuntime()");
		^

Error: Chunk.entry was removed. Use hasRuntime()
    at Chunk.get entry [as entry] (/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/webpack/lib/Chunk.js:59:9)
    at /Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/extract-text-webpack-plugin/index.js:201:13
    at Array.filter (native)
    at Compilation.<anonymous> (/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/extract-text-webpack-plugin/index.js:200:37)
    at Compilation.applyPlugins0 (/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/tapable/lib/Tapable.js:68:14)
    at Compilation.seal (/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/webpack/lib/Compilation.js:558:8)
    at /Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/webpack/lib/Compiler.js:474:16
    at /Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/tapable/lib/Tapable.js:225:11
    at _addModuleChain (/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/webpack/lib/Compilation.js:472:11)
    at processModuleDependencies.err (/Users/markbates/Dropbox/development/gocode/src/github.com/markbates/coke/node_modules/webpack/lib/Compilation.js:443:13)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

stanislas-m:
https://github.com/webpack/webpack/issues/2764

> [...] as others in this thread have pointed out, you need to install a webpack2.0-compatible version of the plugin.
> `npm install extract-text-webpack-plugin@^2.0.0-beta --save-dev`

markbates:
[...]

PR #263, opened 2017-02-27
Webpack v2.2.1 upgrade

fooflare:
As it was issued in #262, buffalo has broken with the last release of extract-text-webpack-plugin. Now it is an stable version (reason to wait in #195) so I decided to upgrade and parse the webpack.config.js to work with webpack v2.2.1 and the last version of the mentioned plugin. I tested it with a generated application including this patch and seems to work fine. I hope to not have forgotten anything.

Solves #262 and #195


markbates:
Woot! Thanks!

fooflare:
A pleasure.

PR #276, opened 2017-03-09
React app template

fooflare:
First try with react, template based on: https://github.com/facebookincubator/create-react-app/tree/master/packages/react-scripts

Buffalo dev and build are working, test needs integration (wip)


fooflare:
Some explanation here. The project structure is a bit different from the other 2 templates, we have `src` folder isntead of `assets` (working folder). `public` now is saved the `index.html` and the favicon file. `build` is the folder where the magic happens, every file is served from this folder. 
Why everything is in `build`? Through webpack, the `index.html` has the css injected in dev mode, and the script src is injected too, so this file have to be built. One thing I'd like to get done is a hot code reload feature (right now I'm just experimenting) in dev mode, I think it'd be great for the future.
In production mode, everything is minified, even the `index.html.` In this case, the css is not injected, but the `script` and link `tags` are. These files are hashed and injected.
One thing I was very careful is to make a postprocessing to not generate new actions files, I just modify the existing ones when are generated: https://github.com/fooflare/buffalo/blob/fa27cea9d850b581b39bb74cbd49e536d42f6c6f/buffalo/cmd/new.go#L193
To generate this template, `--with-react` flag has to be used and yarn is also optional for this template to keep the default package manager as in the rest of the generated applications.
Any questions, suggestions and comments are welcome.

markbates:
I'm curious as to why we can't use `assets/` instead of `src/` to keep things consistent.

fooflare:
We could, in react is generally done with that structure. Of course everything can be configured (not too difficult to change), but the only important thing is to avoid css, js and images folder in assets. That's because when you create a component, it would be tedious to separate css and js files of each one. I just did it like that because it is how I am used to, but as I said, it can be changed and done more like buffalo style. It's just changing public to templates, src to assets and build to public.

fooflare:
New commit to keep assets, public and templates folders structure in the generated application.
Do you feel comfortable with this template? If not, I would understand it and just keep it in my fork. Sometimes some features complicate the maintenance and I should have asked this earlier.

markbates:
I would like to put this up for a vote, if that's alright with you?

I'm not against it, but I would like to know what others thing of having this in Buffalo.

markbates:
Also, I'm working why under `templates` there is a `assets/images/favicon.ico` file. Was that put there by accident?

apaganobeleno:
I think by providing assets compilation/transpilation/serving should continue being part of the buffalo's responsibility, its great to give support to JS frameworks but given there are too many options i think this could be out of buffalo's scope.

IMHO this kind of framework specifics should be built on top of webpack and buffalo's responsibility should be to provide the foundations to integrate react/angular2/ember/...

In other words, i don't think this is (right now) something that should be inside buffalo, seems to highly specific to me (even given react is a great framework).

fooflare:
@apaganobeleno I agree this is maybe out of buffalo's scope, even after finishing it I had doubts about it, mainly because this can be kept separated from buffalo and just explain how to configure it by hand (not too hard for the user).
@markbates the favicon came with the original react template and instead of removing it, I just changed it with a buffalo one, I didn't think about it :).
My opinion is I can make an independent project to keep this integration (not so integrated) working but configuring it by hand with simple steps. I think not so many people have this kind of needs or maybe they would prefer another option. Even keeping it separated, I could afford the generation of angular and polymer templates too (not so sure about this, I have to think about it).
So definitely, I let this decision in your hands @markbates but for me is fine if you put this up for vote or just close it, I think keeping it separated is a good option and less painful for buffalo maintenance.

Thank you both for your comments and for making me see it from another point of view.

jbarone:
Instead of hard coding for specific front ends, would it be possible to think more in terms of dynamic templates. Kinda like what [cookiecutter](https://github.com/audreyr/cookiecutter) does? And possibly let buffalo dynamically enable the generators based on a config. Possibly pie in the sky thinking but with all the different frontends (React, AngularJS, Vue, etc...) doesn't make sense to bind to just one.

markbates:
@fooflare Sorry for dropping the ball on this, my bad. I'm going to close this without merging it. While I like the idea, it just doesn't quite fit the project. When Go releases plugins for all platforms, not just Unix, Buffalo will start using them, and this is the sort of thing that would be a great generator plugin.

In the meantime would you be interesting in writing up how to use React with Buffalo on the blog? http://blog.gobuffalo.io it's a Medium site, so it's really easy to author on.

Thanks so much for taking the time on this. Sorry it's not getting merged.

fooflare:
Don't worry, I completly understand it. I'd love to write an article about it. But now, I have the intention to create a project to generate templates
 around buffalo. I'd like to generate different frontend
 templates. If you want, we can
 keep in touch about this and see how to cover different frontend templates. Of course when plugins are a real option, I'll have a look at it.

markbates:
That would be awesome! Let me know how you get one with it

PR #305, opened 2017-03-20
Fix #181: generate default gitlab ci config

### LINKED ISSUE #181: Generate a default .gitlab-ci.yml file for new projects
See #36. This is the same case, but with a different provider.

stanislas-m:
Still one question remaining:

> My problem is the way the database.yml is generated:
> 
> ```yaml
> test:
>   dialect: postgres
>   database: buffalo-test_test
>   user: postgres
>   password: postgres
>   host: 127.0.0.1
> ```
> 
> Since the PostgreSQL server is not on 127.0.0.1 in this case (it will be on a dynamic host created by Gitlab), the test will fail. So, we can either:
> - ask the user to change the host by hand
> - let Buffalo customize it by tweaking Pop
> - change Pop template to use `envOr` helper on test case (so we can provide env in the ci script)
> - tweak it at build time, with a bash command like `sed` or `awk`
> 
> In my opinion, the third option is the best and would not break legacy setups.


markbates:
"change Pop template to use envOr helper on test case (so we can provide env in the ci script)" 👍

stanislas-m:
@markbates It should work now. I've added a case without Pop too.

markbates:
LGTM. :) Thanks!

Issue #297, opened 2017-03-18
Add basic functions into the default implementation of resource functions


as27:
The generator for resources should generate some more code into the default implementations. The command `buffalo g r users` should add a basic functionality for the model `user`. 

The generated code should be similar like in the [html-resource example](https://github.com/gobuffalo/buffalo/blob/master/examples/html-resource/actions/users.go). But I would suggest not to use a middleware inside the generated code, because of readability issues.

markbates:
It'll be cleaner if you use the new(ish) ValidateAnd... methods https://godoc.org/github.com/markbates/pop#Connection.ValidateAndCreate

markbates:
This probably better code to model after https://github.com/gobuffalo/toodo/blob/master/actions/todos.go

as27:
Inside the documentation it is not defined how to use the resources generator for longer named models.

For one word models that is clear, but whats with longer model names. For example a model `ContentSection`. 

How should the user generate that resource?

```
buffalo g r content_sections
buffalo g r contentSections
buffalo g r ContentSections
```

or should the generator support all possibilities?

markbates:
I would think it should be able to support any of those possibilities.

as27:
I am moving forward. Now I need feedback, because I am not sure if I am on the right track. The following file is completely generated via `buffalo g r pages`:
[Link to the gist: pages.go](https://gist.github.com/as27/c5bf31a0f3e8faef71044d44ee870028)

Not all functions are completed so the status is WIP. The code compiles and works, when the template files exist.

I added some comments to make that code easier to edit. Is this too much? or should there be much more comments? 

Something different: I don't know how to solve this:
In line 32 the path is in [] brackets:
`// Show gets the data for one page. This function is mapped to the path GET /pages/[page_id]`

That is because inside the template following does not work
```
GET /{{under}}/{{{singular}}_id} <-- causes an error
GET /{{under}}/\{{{singular}}_id} <-- renders as GET /pages/{{{singular}}_id}
```

markbates:
Wow! This is looking really good so far! Very impressed!

To answer the last question first, you need to escape the `{{` otherwise Go templates tries to execute it. Here's an example of how to escape it:

https://github.com/markbates/pop/blob/master/soda%2Fcmd%2Fgenerate%2Fconfig_templates.go#L12

It's ugly, but that's just how you do it. :(

I'm all for comments in the generated code, the more the merrier!

[...]

PR #313, opened 2017-03-23
Resource generator template #297

as27:
Added basic implementations for the resource functions.


markbates:
Looks like the tests are failing. :(

You can run the CI tests locally with `docker build .` so you don't have to keep committing and pushing to see if they've passed.

as27:
My mistake I didn't check the tests. But this is maybe deeper, because the generated file expects that there is a type inside the models package. So if I generate for users and there is no model for it the code does not compile. This will need some time for me to fix.

PR #315, opened 2017-03-24
Fix render error with rice

XelaRellum:
This error happens only if rice is used to actually create rice-box.go.

Rice expects paths separated by slashes, but filepath.Join produces backslashes on Windows.

### DIFF
--- a/render/template.go
+++ b/render/template.go
@@ -33,15 +33,25 @@ func (s templateRenderer) Render(w io.Writer, data Data) error {
 	w.Write([]byte(body))
 	return nil
 }
+
+// FilepathJoin works just like filepath.join but will always use slashes.
+// Note: on Windows filepath.join returns a path separated by backslashes.
+// While this works find with files it crashes with the use of rice which
+// expects slashes
+func FilepathJoin(dirname string, filename string) string {
+	joined := filepath.Join(dirname, filename)
+	return strings.Replace(joined, "\\", "/", -1)
+}
+
 func (s templateRenderer) partial(name string, dd Data) (template.HTML, error) {
 	d, f := filepath.Split(name)
-	name = filepath.Join(d, "_"+f)
+	name = FilepathJoin(d, "_"+f)
 	return s.exec(name, dd)
 }
 
 func (s templateRenderer) exec(name string, data Data) (template.HTML, error) {
 	var body string
-	source, err := s.Resolver().Read(filepath.Join(s.TemplatesPath, name))
+	source, err := s.Resolver().Read(FilepathJoin(s.TemplatesPath, name))
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


markbates:
So in the `v0.8.0` branch, which will be released on Monday, we have removed Rice (and the `resolvers` package) entirely from Buffalo. We're switching to http://github.com/gobuffalo/packr instead. It would be great to check if this is a problem in that branch. I don't have Windows, so I can't easily reproduce.

I'm going to close this PR since Rice support is going to moot after Monday. Thanks.

XelaRellum:
Ok, will try. I was starting to use go-bindata for my project, but then
switched over to rice since support for that was already there...

Well, just checked
https://github.com/gobuffalo/packr/blob/master/box.go

filepath.Join is used in several places. Will break in the same places
since filepath.Join uses backslashes as dir separator in Windows.


Alex

Am 24.03.2017 um 18:54 schrieb Mark Bates:

XelaRellum:
So I tried to check out v0.8.0 and as guessed it doesn't work on Windows. The generated file doesn't compile (since it contains backslashes for directories) which I tried to fix but then the files are not found. Please consider re-opening this pull request and create a bugfix release v0.7.4.2 so I can stick with rice for the time being until packr works on Windows. Thanks!

markbates:
I guess I really don't understand what is happening here. Can you explain it better?

XelaRellum:
Ok. Here is an extract of the generated file:
```go
package actions

import "github.com/gobuffalo/packr"

// !!! GENERATE FILE !!!
// ...
func init() {
	packr.PackJSONBytes("../templates", "..\templates\_navbar.html", "\" .....")
	packr.PackJSONBytes("../templates", "..\templates\article.html", "\"PHVsI... ")
	packr.PackJSONBytes("../templates", "..\templates\articleattributes.html", "\"PHVsIG .... ")
}
```

As you can see the generated code causes errors since the paths are separated by unescaped backslashes. The canonical form of a path (using slashes) can be create by filepath.ToSlash ...

markbates:
There was a lot of activity this weekend on packr to make it work with windows. Can you try the latest version of it and see if it solves your problem?

XelaRellum:
Updated packr to the latest version and generated code seems to be okay.
Right now I don't see anything, but thats because the templating engine
seemed to have changed from velvet to plush (with a different syntax).

Am 03.04.2017 um 12:51 schrieb Mark Bates:

markbates:
It has changed, but you can change it back to Velvet if you want https://blog.gobuffalo.io/buffalo-v0-8-0-released-33f1786eadf

Personally, I would take the time to upgrade the templates, but switching the template engine to Velvet should make it work.

XelaRellum:
Okay, migrated to plush. What I liked with buffalo 0.7.x better was the independence on a thing like packr or rice as long as it fulfilled the http.FileSystem interface. With 0.8.x we now depend on packr. Would be besser if it depended on an interface than on a concrete type.

PR #323, opened 2017-03-25
fix output path

mario-seidel:
otherwise webpack will not fill that dir with files

os: linux, wepack version: 2.3.2


markbates:
Thanks.

PR #329, opened 2017-03-26
Adds a buffalo version command to the CLI

SpencerCDixon:
I had a little trouble getting dev environment set up and was wondering if anyone could provide some assistance.  Slash might be nice to have a "How to contribute" doc somewhere?

After forking the repo I `git cloned` it into my go github.com/spencercdixon workspace.  When I tried to build the CLI though I get this:

screen shot 2017-03-26 at 10 16 21 am

This was the first Go project I ever worked on that wasn't a "hello world" type thing though so I'm probably just missing some very obvious steps haha.

Also, does go build cache the last successful build or something?  If I cd'ed into `buffalo/cmd` and ran `go build` it would always build the old version of buffalo even after I checkouted the latest 0.8 branch.  I feel like that's not normal....?

### DIFF
--- a/buffalo/cmd/version.go
+++ b/buffalo/cmd/version.go
@@ -1,4 +1,23 @@
 package cmd
 
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
 // Version is the current version of the buffalo binary
 const Version = "0.8.0.dev"
+
+func init() {
+	RootCmd.AddCommand(versionCmd)
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print the version number of buffalo",
+	Long:  `All software has versions.  This is buffalo's.`,
+	Run: func(c *cobra.Command, args []string) {
+		fmt.Println("Buffalo version %s", Version)
+	},
+}


markbates:
Here is a good article on how to contribute to Go projects. The problem you are having is due to namespaces. This article spells out exactly how to do it. It's actually very straightforward, but not intuitive at all. 

https://splice.com/blog/contributing-open-source-git-repositories-go

markbates:
When you run `go build` it builds the binary in that folder, but doesn't install it. 

If you run `go install ./buffalo from the root of the Buffalo project it will build and install the binary.

buffalo/cmd/version.go
```diff
@@ -1,4 +1,23 @@
...
+func init() {
+	RootCmd.AddCommand(versionCmd)
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print the version number of buffalo",
+	Long:  `All software has versions.  This is buffalo's.`,
+	Run: func(c *cobra.Command, args []string) {
+		fmt.Println("Buffalo version %s", Version)
```
markbates: I don't think this will do what you want it to do. It should probably be:

```go
fmt.Printf("Buffalo version %s\n", Version)
```

SpencerCDixon:
Great thank you so much! Will read through this and get this fixed up 👍

PR #382, opened 2017-04-12
Fixes nested resources generation

### LINKED ISSUE #364: Resource generator doesn't support "nested" values
```
buffalo g resource admin/planes name
Buffalo version 0.8.1.dev

--> actions/admin/planes.go
--> actions/admin/planes_test.go
--> locales/admin/planes.en-us.yaml
--> templates/admin/planes/_form.html
--> templates/admin/planes/edit.html
--> templates/admin/planes/index.html
--> templates/admin/planes/new.html
--> templates/admin/planes/show.html
--> buffalo db g model admin/plane name
v3.13.7

--> models/admin/plane.go
--> models/admin/plane_test.go
--> goimports -w .
actions/admin/planes.go:22:11: expected type, found '/'
actions/admin/planes.go:38:2: expected declaration, found 'if'
actions/admin/planes_test.go:9:34: expected '(', found '/'
actions/admin/planes_test.go:10:2: missing ',' in parameter list
actions/admin/planes_test.go:11:1: expected 'IDENT', found '}'
actions/admin/planes_test.go:13:1: missing ',' in parameter list
actions/admin/planes_test.go:17:34: expected '(', found '/'
actions/admin/planes_test.go:18:2: missing ',' in parameter list
actions/admin/planes_test.go:19:1: expected 'IDENT', found '}'
actions/admin/planes_test.go:21:1: missing ',' in parameter list
actions/admin/planes_test.go:25:34: expected '(', found '/'
actions/admin/planes_test.go:26:2: missing ',' in parameter list
actions/admin/planes_test.go:27:1: expected 'IDENT', found '}'
models/admin/plane.go:12:11: expected type, found '/'
models/admin/plane.go:22:2: expected declaration, found 'return'
models/admin/plane.go:26:11: expected type, found '/'
models/admin/plane.go:31:2: expected declaration, found 'return'
models/admin/plane_test.go:5:16: expected '(', found '/'
```

apaganobeleno:
@markbates this one fixes #364


markbates:
I'm not near a computer, but I know that there's more to this. :)

For example, in my example there's a new `admin` package created, which is where the generated resource lives. That packages doesn't have a render engine or an action suite like th `actions` package does, which means the code won't compile. 

This needs to address both the fact that these might or might not exist in the package. 

Also, models shouldn't be generated in subsequent packages but rather in the `models` package.

apaganobeleno:
Oh! yikes! you're right, too easy to be true, i have the second part covered, will continue looking into the rest and keep you posted.

apaganobeleno:
hi @markbates i have it working with the following code : 

1. Adding a render file into the nested package that looks like this:
```go
package {{.package}}

import (
)

var r *render.Engine

// SetRenderEngine sets the r variable from outside (p.e. actions root) package.
func SetRenderEngine(engine *render.Engine) {
	r = engine
}
```

2. Setting the render engine from the app.go like:

```go
admin.SetRenderEngine(r)
app.Resource("/planes", admin.PlanesResource{&buffalo.BaseResource{}})
```

WDYT about this approach ?

apaganobeleno:
@markbates LMK when you have the chance to see this one, will love to read your feedback on this approach.

markbates:
I've been thinking about this ticket all weekend. I think the way to solve it is this:

https://github.com/gobuffalo/buffalo/issues/394

Thoughts?

apaganobeleno:
I agree on moving into config, that was my second option to solve this, but at that time i was working on this i thought on backwards compatibility as a MUST, i think one thing to keep in mind with this change is some sort of migration guide so, old versioned (beta) app builders can move to 0.9.0.

Overall this provides a level of separations that will allow us to support nested resources without generating much (maybe duplicated) code, i agree on this one.

apaganobeleno:
@markbates i'm closing this one given we're going to attack this feature in combination with the new  config folder.

PR #390, opened 2017-04-16
Added XML resource generator, fixes #388

### LINKED ISSUE #388: Resource generator should be able to generate XML
Since the JSON resource generator is working now, we should do the same for its XML friend. :)

stanislas-m:


markbates:
@stanislas-m this looks good, but the validation library doesn't support XML encoding, so I need to fix that first before I merge this. As soon as I do, this will be merged.

PR #400, opened 2017-04-20
add forbidden app names validation check

marythought:
Add name validation so a user can't name a project "buffalo" (which will break the binary and cause headaches). It's designed to check against a slice of strings so that if there are other namespaces to protect, those can easily be added in the future.


markbates:
@marythought can you please change the branch from `master` to `development`? GitHub is acting up and won't let me do it on my phone. :) Thanks.

markbates:
Thanks Mary!!

PR #406, opened 2017-04-28
Added a ldflags option to the build command

tbasnett:
During our deployment process I was needing to pass my own -ldflags options through to the go build command.

So I've added the same option to the buffalo build command and pass them down. Throwing an error if it matches the '-X main.version=' or '-X main.buildTime=' options which are used by the build command.


buffalo/cmd/build.go
```diff
@@ -4,12 +4,14 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
+	"errors"
```
markbates: this line is causing the tests to fail, there is already an errors package declared below.

buffalo/cmd/build.go
```diff
@@ -401,6 +421,7 @@ func init() {
 	buildCmd.Flags().StringVarP(&buildTags, "tags", "t", "", "compile with specific build tags")
 	buildCmd.Flags().BoolVarP(&zipBin, "zip", "z", false, "zips the assets to the binary, this requires zip installed")
 	buildCmd.Flags().BoolVarP(&extractAssets, "extract-assets", "e", false, "extract the assets and put them in a distinct archive")
+	buildCmd.Flags().StringVarP(&ldflags, "ldflags", "", "", "set any ldflags to be passed to the go build")
```
markbates: You can use `buildCmd.Flags().StringVar` if you're not using a short-hand flag.

buffalo/cmd/build.go
```diff
@@ -343,7 +346,24 @@ func (b *builder) buildBin() error {
...
+
+	flags := []string{
+		fmt.Sprintf("-X main.version=%s", version),
+		fmt.Sprintf("-X main.buildTime=%s", buildTime),
+	}
+
+	// Add any additional ldflags passed in to the build args
+	if len(ldflags) > 0 {
+		if foundVersion, _ := regexp.MatchString("-X\\s+main.version=", ldflags); foundVersion {
+			return errors.New("The ldflag option '-X main.version=' is reserved for gobuffalo use")
```
markbates: Go prefers errors to start with a lower case. Also please change `gobuffalo` to `Buffalo`.

buffalo/cmd/build.go
```diff
@@ -343,7 +346,24 @@ func (b *builder) buildBin() error {
...
+		}
+		if foundBuildTime, _ := regexp.MatchString("-X\\s+main.buildTime=", ldflags); foundBuildTime {
+			return errors.New("The ldflag option '-X main.buildTime=' is reserved for gobuffalo use")
```
markbates: Go prefers errors to start with a lower case. Also please change `gobuffalo` to `Buffalo`.

markbates:
@tbasnett the tests are failing, and there a few minor tweaks that need to be made, but otherwise, looks good. it'll be nice to have.

tbasnett:
Thanks! I'm going to close this and create a new PR since I realized I was in the master branch instead of developer.

PR #409, opened 2017-04-28
Added a ldflags option to the build command

tbasnett:
Added fixes suggested in here: https://github.com/gobuffalo/buffalo/pull/406. Also, made the pr from the correct branch :)

During our deployment process I was needing to pass my own -ldflags options through to the go build command.

So I've added the same option to the buffalo build command and pass them down. Throwing an error if it matches the '-X main.version=' or '-X main.buildTime=' options which are used by the build command.


markbates:
Awesome! Thanks.

PR #410, opened 2017-04-30
WIP: Add auto mode to resource generator

stanislas-m:
Using auto mode, the generated resource will be able to produce XML or
JSON contents, according to the "Content-Type" HTTP header, or the
"format" query argument.

If the content type can't be determined by either the HTTP header or the
"format" query argument, it fallbacks to JSON.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,6 +67,13 @@ RUN rm models/user_test.go
 RUN rm models/user.go
 RUN rm actions/users_test.go
 
+RUN buffalo g resource --type=auto users name:text email:text
+RUN filetest -c $GOPATH/src/github.com/gobuffalo/buffalo/buffalo/cmd/filetests/resource_json-xml.json
+
+RUN rm models/user_test.go
+RUN rm models/user.go
+RUN rm actions/users_test.go
+
 WORKDIR $GOPATH/src
 RUN buffalo new --skip-pop simple_world
 WORKDIR ./simple_world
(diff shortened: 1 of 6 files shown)


markbates:
I've been thinking about this one over the weekend. I wonder if we can also include `HTML` rendering in there, and make that the default. 

Here's what I'm thinking, we know the HTTP method and the URL. With that we can make some assumptions:

| METHOD | URL | PATH|
|---|---|---|
| GET | /users | users/index.html |
| GET | /users/new | users/new.html |
| POST | /users | users/new.html |
| GET | /users/1 | users/show.html |
| GET | /users/1/edit | users/edit.html |
| PUT | /users/1 | users/edit.html |
| DELETE | /users/1 | users/show.html |

Obviously, if people change things, don't want to use the defaults, etc... they have to handle the rendering more manually, but if you working on a "standard" resource then these conventions should be pretty easy to map and handle.

The following would then be able to handle HTML, JSON, and XML:

```go
c.Render(200, r.Auto(user, c.Request())
```

stanislas-m:
I don't think the PATH is relevant for PUT, POST and DELETE, but an auto handler can be pretty cool, yeah!

markbates:
It actually is relevant. Think of an error case:

```go
if verrs.HasAny() {
  return c.Render(422, c.Auto(user, c.Request())
}
```

stanislas-m:
Oh, I see. Maybe, the DELETE case is not relevant, though.

PR #421, opened 2017-05-05
RequestLogger sets request_id in context

nzoschke:
Set the logger request_id in the context, which enables the rest of the request handling chain to use the ID. 

This is useful for:

- Including the request ID on errors sent to a service like Rollbar
- Passing the request ID to a subsystem to correlate logs


markbates:
Thanks.

PR #422, opened 2017-05-06
Change default host from 127.0.0.1 to localhost

samus:
This is a minor change to make development testing of oAuth requests easier.  What I found was that I would load up the local server through http://localhost/auth/discord and after granting access the call back uri was http://127.0.0.1/auth/discord/callback. This would cause buffalo to throw a 401 error saying the session couldn't be found.

### DIFF
--- a/options.go
+++ b/options.go
@@ -28,7 +28,7 @@ type Options struct {
 	// SessionName is the name of the session cookie that is set. This defaults
 	// to "_buffalo_session".
 	SessionName string
-	// Host that this application will be available at. Default is "http://127.0.0.1:[$PORT|3000]".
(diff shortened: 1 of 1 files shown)


markbates:
I appreciate this PR, but I'm not merging it. The default is actually set to `127.0.0.1` *because* of this reason. Some OAuth providers, GitHub, for example, don't allow the host to be set to `localhost`, but they *all* allow `127.0.0.1`, so it makes the most sense to have that as the default.

Perhaps this PR would be best modified into a documentation PR or, maybe an improved error message at http://github.com/markbates/goth?

I hope this makes sense.

PR #435, opened 2017-05-13
Fix HTTP Basic browser authentication prompt

peteretelej:
Add WWW-Authenticate header and 401 response status to make browsers prompt entry of Basic authentication username and password.


markbates:
Yay! Thanks.

PR #442, opened 2017-05-18
cleaned up some non-idiomatic code

markbates:


apaganobeleno:
@markbates do you by chance know why codeclimate is not reporting ?

markbates:
No idea. I'll look into when I get back stateside.

PR #443, opened 2017-05-21
fixed `http: multiple response.WriteHeader calls`

jnoortheen:
I was getting this error message printed out in my logs. Removing one line from errors.go fixed that.


markbates:
Thanks! This has bothered me for ages, but I kept forgetting to figure out what it was.

PR #446, opened 2017-05-25
generate a Dockerfile and .dockerignore file for new apps

markbates:


generators/newapp/templates/dot-dockerignore.tmpl
```diff
@@ -0,0 +1,3 @@
+node_modules/
+*.log
+bin/
```
apaganobeleno: @markbates should we add `.git` here ?

markbates: I originally had that, but then the when building the binary we no longer have the git information needed to build the `ldflags`, so the SHA doesn't get burnt into the binary. People can add `.git`, and anything else they want, this is meant to just be a "starter" file.

apaganobeleno: Another thing we could consider is using a [multi-stage Dockerfile](https://docs.docker.com/engine/userguide/eng-image/multistage-build/#before-multi-stage-builds) that will end up with a slimmer docker image, but i think is ok to leave it for the future.

Issue #366, opened 2017-04-06
Resource generator should generate default tests


markbates:
Right now it generates tests that look like this:

```go
func (as *ActionSuite) Test_BooksResource_List() {
	as.Fail("Not Implemented!")
}
...
```

It would be fantastic if these tests were actually implemented when generating a new resource.

This ticket is not for the faint of heart. :)

apaganobeleno:
@markbates i can take this one :)

as27:
@apaganobeleno if you need some help, just let me know.

apaganobeleno:
@as27 would you like to take this one? i'm afraid i wont have much time for the next month and don't want to delay this more.

as27:
@apaganobeleno yes of course. I would love to help out here.

as27:
@markbates I would make the tests analog to the todo app. The tests will also use `ActionSuite` like here:
https://github.com/gobuffalo/toodo/blob/master/actions/todos_test.go

markbates:
That makes sense.

as27:
The test generator is almost ready. But when I run the tests I receive always return code 500. I think I need a little help to understand why.

The template file for the generator would be here:
https://github.com/as27/buffalo/blob/resource-test-generator/generators/resource/templates/actions/resource-use_model_test.go.tmpl

I made a own repository for a generated example. The output after code generation is:
https://github.com/as27/buffaloresourcetest/blob/master/actions/users_test.go

But when I run `buffalo test` inside this repository I get always 500 as return code.

stanislas-m:
@as27 Giving a quick look, it seems you have a typo in https://github.com/as27/buffalo/blob/resource-test-generator/generators/resource/templates/actions/resource-use_model_test.go.tmpl line 49

`{{.Singular}}` -> `{{.singular}}`

The other errors seem linked to the lack of CSRF token in your request.

as27:
@stanislas-m thanks. CSRF was the missing link I didn't see. But that issue is more a `gobuffalo/suite` issue then a generator thing ([look here](https://github.com/gobuffalo/suite/blob/master/suite.go#L41))? Or am I wrong?

stanislas-m:
@as27, as I understand the way the CSRF is handled in the test suite, it just replaces the middleware with a no-op. Maybe I'm wrong, but I think the `app.Use` gets a copy of the middleware, so, if you replace it after the `app.Use` call it can't work.

[...]

PR #449, opened 2017-05-29
Resource default test generator #366

as27:


markbates:
Need to remove `buffalo/buffalo2` from the PR

markbates:
```bash
$ buffalo g resource post title body:nulls.Text


actions/posts_test.go:31: cannot use "A string for Body" (type string) as type nulls.String in field value
actions/posts_test.go:35: cannot use "Another string for Body" (type string) as type nulls.String in field value
actions/posts_test.go:70: cannot use "A value for Body" (type string) as type nulls.String in field value
actions/posts_test.go:115: cannot use "A value for Body" (type string) as type nulls.String in field value
actions/posts_test.go:137: cannot use "Initial value for Body" (type string) as type nulls.String in field value
actions/posts_test.go:147: cannot use "Changed value for Body" (type string) as type nulls.String in field value
```

markbates:
We need to do a better job of filling in good random values for the models.

as27:
> We need to do a better job of filling in good random values for the models.

@markbates yes you are right. 
But to handle all the logic inside the template is too much. I am going to write a helper for the template. First I would add this new helper inside `generate/resource.go` via `makr.Helper["createDefaultValue"] = ...`  after that the definition could move into `makr/helpers.go` or better into `pop`, because when there will be a new type inside `pop` no other package had to be changed. What do you think?

markbates:
I'm thinking that a good "first pass" would be a comment above the struct in the test that says something like:

```go
func (as *ActionSuite) Test_TodosResource_List() {
	// TODO: fill-in "models.Todo{}" with appropriate test data
	todos := models.Todos{}
	for _, t := range todos {
		err := as.DB.Create(&t)
		as.NoError(err)
	}

	res := as.HTML("/todos").Get()
	body := res.Body.String()
	for _, t := range todos {
		as.Contains(body, fmt.Sprintf("<h2>%s</h2>", t.Title))
	}
}
```

or we could use a failure message:

```go
func (as *ActionSuite) Test_TodosResource_List() {
	as.FailNow(" TODO: fill-in 'models.Todo{}' with appropriate test data")
	todos := models.Todos{}
	for _, t := range todos {
		err := as.DB.Create(&t)
		as.NoError(err)
	}

	res := as.HTML("/todos").Get()
	body := res.Body.String()
	for _, t := range todos {
		as.Contains(body, fmt.Sprintf("<h2>%s</h2>", t.Title))
	}
}
```

That would be "good enough" for a first pass. 

As a second pass, I think it would be great to either find, or build, a tool that will take a struct (or slice) and generate random values for it's data.

```go
func (as *ActionSuite) Test_TodosResource_Create() {
	todo := &models.Todo{}
	somepackage.Fill(todo, somepackage.Options{
		"ID": {Skip: true},
		"Body": {Filler: func() interface{} {
			return "some body text"
		}},
	})
	/*
	values would be:
	todo.ID = ""
	todo.Body = "some body text"
	todo.Title = "asdfkjasdfjasdfkjsda"
	*/
	res := as.HTML("/todos").Post(todo)
	as.Equal(301, res.Code)
	as.Equal("/todos", res.Location())

	err := as.DB.First(todo)
	as.NoError(err)
	as.NotZero(todo.ID)
	as.NotZero(todo.CreatedAt)
	as.Equal("Learn Go", todo.Title)
}
```

markbates:
Just started playing with it. So far, looks good. Noticed that there is a missing `.` in the error messages:

```
TODO: fill-in 'modelsPost{}' with appropriate test data
```

PR #469, opened 2017-06-14
Nested resources with actions namespacing

apaganobeleno:
This PR covers #364


markbates:
Route names aren't being generated correctly, and the generated templates don't have the correct routes.

`$ buffalo g admin/user name bio:text`

http://localhost:3000/admin/users/new:

usersPath() (reflect.Value) is an invalid function error inside of calling function

```
METHOD	PATH	NAME	HANDLER
GET	/	rootPath	github.com/markbates/coke/actions.HomeHandler
GET	/admin/users	usersAdminPath	github.com/markbates/coke/actions.AdminUsersResource.List
POST	/admin/users	usersAdminPath	github.com/markbates/coke/actions.AdminUsersResource.Create
GET	/admin/users/new	newUserPath	github.com/markbates/coke/actions.AdminUsersResource.New
GET	/admin/users/{user_id}	userAdminPath	github.com/markbates/coke/actions.AdminUsersResource.Show
PUT	/admin/users/{user_id}	userAdminPath	github.com/markbates/coke/actions.AdminUsersResource.Update
DELETE	/admin/users/{user_id}	userAdminPath	github.com/markbates/coke/actions.AdminUsersResource.Destroy
GET	/admin/users/{user_id}/edit	editUserPath	github.com/markbates/coke/actions.AdminUsersResource.Edit
```

markbates:
I'm thinking this will have to wait to 0.9.1. I don't feel comfortable with getting it fully tested before tomorrow. You ok with that?

apaganobeleno:
@markbates agree with you, more given we don't know if this is something most of the people uses as a basis, in my case i don't usually do nested resources so i think this can wait for 0.9.1.

On the other side i think is all-set to be merged so, we can regroup after 0.9.0 to review if its all set or iterate on it.

apaganobeleno:
@markbates please take a look at this one, i think is ready for review. (again :D)

markbates:
Ok. I'll review it this afternoon.

-----------
Mark Bates

markbates:
@apaganobeleno this looks good, only problem is that the templates are still using the wrong paths. For example it's still generating `newUsersPath()` instead of `newAdminUsersPath()`.

apaganobeleno:
Got it, will take care of these, thanks Mark, and sorry about too many back and forth, i've been thinking we should have some more integration tests that reduces the testing effort before accepting a PR on the generators, anyways, that still in my brain's background thread for things we do in the future to enhance the process.

apaganobeleno:
@markbates i just checked and it seems to be doing it correctly, would it be ok to add some filetests  that ensure the correct route being used as a way to demonstrate/ensure-for-future this is covered?

markbates:
Yay!! Nice work on this one!

Issue #477, opened 2017-06-21
Grifts generator


apaganobeleno:
It would be good to have a generator for grift tasks.

CLI:

```sh
$ buffalo g grift taskName
```

Would generate:
```
grifts/task_name.go
```

And contents of the grift task would be something like:

```go

package grifts

import (
	. "github.com/markbates/grift/grift"
)

var _ = Desc("task_name", "TODO")
var _ = Add("task_name", func(c *Context) error {
	return nil
})

```

cc @markbates

markbates:
I would also generate a `Desc` function as well.

```go
package grifts

import (
	. "github.com/markbates/grift/grift"
)

var _ = Desc("task_name", "TODO")
var _ = Add("task_name", func(c *Context) error {
	return nil
})
```

markbates:
Also, if the task name contains `:` we should use `Namespace`:

```bash
$ buffalo g grift a:b:c
```

```go
package grifts

import (
	. "github.com/markbates/grift/grift"
)

var _ = Namespace("a", func() {
  Namespace("b", func() {
    Desc("c", "TODO")
    Add("c", func(c *Context) error {
	return nil
    })
  })
})

```

apaganobeleno:
Nice!! i didn't know about namespaces in grift! thanks @markbates

markbates:
@apaganobeleno that's because it only got merged in today! :) https://github.com/markbates/grift/pull/10

apaganobeleno:
covered in #479

PR #479, opened 2017-06-21
Feature grift generator

apaganobeleno:
Adds grift generator mentioned in #477


buffalo/cmd/generate/grift.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/gobuffalo/buffalo/generators/grift"
+	"github.com/gobuffalo/makr"
+	"github.com/markbates/inflect"
+	"github.com/spf13/cobra"
+)
+
+//GriftCmd is the command called with the generate grift cli.
+var GriftCmd = &cobra.Command{
+	Use:     "grift [name]",
```
markbates: Let's change `Use` to `task [name]` and keep the aliases as they are.

buffalo/cmd/generate/grift.go
```diff
@@ -0,0 +1,50 @@
...
+			return errors.New("you need to provide a name for the grift tasks")
+		}
+
+		var parts []string
+		plain := strings.Contains(args[0], ":") == false
+		filename := fmt.Sprintf("%v.go", inflect.Underscore(args[0]))
+
+		if !plain {
+			parts = strings.Split(args[0], ":")
+			filename = fmt.Sprintf("%v.go", inflect.Underscore(parts[0]))
```
markbates: file name should be the last part, not the first. so `buffalo g task db:seed` should generated `grifts/seed.go`

markbates:
Another thing:

```bash
$ buffalo g grift my:nested:task:is:awesome
$ buffalo g grift my:nested:task
```

both overwrite the same file. we should probably check if the file already exists and append to it instead of writing to it.

other than those few things, this looks pretty good. good job on handling `n` level namespacing.

apaganobeleno:
thanks for the feedback! i'll work on that and keep you posted

apaganobeleno:
@markbates what about using the parts on the namespace to generate diferent files ? :)

```
buffalo g grift my:grift 
- my_grift.go
buffalo g grift my:grift:is:better
- my_grift_is_better.go
```

apaganobeleno:
(i'm working on appending the new task meanwhile)

apaganobeleno:
Ok just covered the case where the file exists, LMK once you have the chance if there are other cases to cover.

generators/grift/grift.go
```diff
@@ -0,0 +1,68 @@
...
+    var _ = Add("{{.taskName}}", func(c *Context) error {
+        return nil
+    })
+{{ else }}  
+    {{ $last := .last }}
+    {{ range $index, $element := .parts }}
+        {{ $isLast := eq $index $last }}
+        {{ if not $isLast }}
+            {{if eq $index 0}}
+                var _ = Desc("{{$element}}", "TODO")
```
markbates: Can you removed the `Desc` for `Namespace` commands? It doesn't do anything, so there's no point in generating them.

buffalo/cmd/generate/grift.go
markbates: Can you change this to `task`?

apaganobeleno: Yes, sure!

buffalo/cmd/generate/grift.go
apaganobeleno: Got it.

markbates:
Looks good. I'm going to merge this. There's just one thing that would be great if it could be improved, but not super critical.

```bash
$ buffalo g task my:cool:task
$ buffalo g task my:other:task
```

generate:

```go
package grifts

import (
	. "github.com/markbates/grift/grift"
)

var _ = Namespace("my", func() {

	Namespace("cool", func() {

		Desc("task", "TODO")
		Add("task", func(c *Context) error {
			return nil
		})

	})

})

var _ = Namespace("my", func() {

	Namespace("other", func() {

		Desc("task", "TODO")
		Add("task", func(c *Context) error {
			return nil
		})

	})

})
```

it would be great if they were in the same namespace, but it's definitely not a show stopper.

apaganobeleno:
@markbates right, i was going to handle this with AST but before i started there was a case i didn't know how to handle:

Lets say you have this in `some.go`

```go

package grifts

import (
	. "github.com/markbates/grift/grift"
)

var _ = Desc("some", "TODO")
var _ = Add("some", func(c *Context) error {
        return nil
})
```

And then you run `buffalo g task some:cool:task`, how would this file end up being ?

PR #480, opened 2017-06-22
Moves destroy model command

apaganobeleno:
Moves destroy model command to be inside `db` commands as the generator for models is inside the same db root-command from pop.


markbates:
@apaganobeleno I'm about to get off an airplane. this looks good. can you merge it into `development` when the tests pass?

apaganobeleno:
all set, thanks Mark!

PR #483, opened 2017-06-26
Fix parseTime to return an error if none format pass

vdemeester:
Before this fix, it will never return an error as the time and error
variable are scoped in the for loop (thus, the upper variables always
have empty values, i.e. nil)

/cc @markbates 

Signed-off-by: Vincent Demeester <vincent@sbr.pm>


markbates:
Thanks!

PR #494, opened 2017-07-06
fix case-insensitive import collision

s7v7nislands:
when multipackage using different versions, go build error:
case-insensitive import collision: "github.com/Sirupsen/logrus" and "github.com/sirupsen/logrus"


markbates:
Thanks!

PR #496, opened 2017-07-06
Add support for plugins for the buffalo binary closes #495

### LINKED ISSUE #495: Add support for plugins for the buffalo binary
It will look in `$PATH` and the `./plugins` directory.

Requirements:
  * file/command must be executable
  * file/command must start with `buffalo-`
  * file/command must respond to `available` and return JSON of `plugins.Commands{}`

Examples:

```ruby
#!/usr/bin/env ruby
#./plugins/buffalo-niles.rb

require 'json'

command = ARGV[0]

case command
when 'available'
  puts JSON.generate([{ name: 'niles', buffalo_command: 'root', description: "he's frasiers brother" }])
when 'niles'
  puts 'niles crane!!!'
end
```

```bash
#!/bin/sh
#./plugins/buffalo-hello.sh

case $1 in
  "available" )
    echo '[{"name":"hello", "buffalo_command": "root", "description": "says hello"}]'
    ;;
  "hello")
    echo "hello world!"
    ;;
esac
```

markbates:
https://vimeo.com/224524780


apaganobeleno:
I'll take a deep looker tonight and add my review, thanks Mark, all this is looking very exciting for Buffalo!

apaganobeleno:
Looks good, i'm merging this one, as @bscott said it would be good to add some blacklist of plugin names that we would like to keep for the future, other than that, thanks Mark for adding me as a reviewer, reading this took me understand our future plugin system, looking forward for writing cool plugins :)

markbates:
We actually don't need to add a black list. Cobra will let you mount commands with the same name, however, it will only run the first one mounted, which will always be our commands. :) I tried it. Works like a should.

PR #511, opened 2017-07-12
Update string and plain renderer with the charset

ryanfaerman:
This enables the plaintext and string renderers to render unicode in the browser.


markbates:
Thanks Ryan!

Issue #508, opened 2017-07-12
Running buffalo commands outside of the directory should error out


corylanou:
Currently you can run commands like "buffalo setup" outside of a project and you will either get no errors, or errors that don't make sense (for obvious reasons).

The binary should verify you are running these commands from within the project otherwise throw an error.

apaganobeleno:
Covered by #514, thanks for adding this one @corylanou

PR #514, opened 2017-07-12
Prevents buffalo from executing commands outside the buffalo project

apaganobeleno:
This will prevent to run buffalo commands outside a buffalo folder by checking `.buffalo.dev.yml` existence.

This doesn't apply for `new` and `version` buffalo's subcommand.


apaganobeleno:
This is related with #508

PR #529, opened 2017-07-17
Use i18n in index template

slomek:
I think that since there is an example i18n message in an initialized project, it would be cool for it to be used in the template. My first step to see how _locales_ work was to search where the key is used, but it wasn't. I needed to go through i18n middleware code to see how I should use it, would be much easier for a less Go-savvy user just to see it in place from the beginning.


markbates:
Good catch. I thought it was already being used there! :)

PR #535, opened 2017-07-18
webpack should rely on webpack config not -p flag

tonyalaribe:
The `buffalo build` command runs webpack with a `-p` flag, which forces tree shaking and default minification configurations on the project.  This can be a problem for projects with more advance settings, with custom requirements. 

This pull request simply moves the same advantages of a `-p` flag, to the webpack.config.js, making it possible for any user to customise the webpack.config.js, as they see fit, without running into similar limitations.


markbates:
Thanks!

PR #537, opened 2017-07-18
added an `info` command to print of env/versions to help with debugging

markbates:


buffalo/cmd/info.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		bb.WriteString("\n### SQLite Version\n")
+		if _, err := exec.LookPath("sqlite3"); err == nil {
+			c = exec.Command("sqlite3", "--version")
+			c.Stdout = bb
+			c.Stderr = bb
+			c.Run()
+		} else {
+			bb.WriteString("MySQL Not Found\n")
```
apaganobeleno: Here should say "SQLite Not Found" instead of MySQL

markbates: fixed

PR #542, opened 2017-07-19
updated to a regex to check for illegal characters

markbates:


markbates:
@apaganobeleno I change the illegal name checker to use a regex so it's more strict.

Issue #503, opened 2017-07-11
Add explicit version (or range) to the Node packages


apaganobeleno:
To ensure better stability with Node packages we need to setup versions for the npm packages we're install when creating a new app, here is the list of packages we currently install, some of them already have versioning others doesn't

- "webpack@~2.3.0"
- "sass-loader"
- "css-loader"
- "style-loader"
- "node-sass"
- "extract-text-webpack-plugin"
- "babel-cli" 
- "babel-core"
- "babel-preset-env"
- "babel-loader"
- "url-loader"
- "file-loader"
- "jquery"
- "bootstrap"
- "path"
- "font-awesome"
- "npm-install-webpack-plugin"
- "jquery-ujs",
- "copy-webpack-plugin"
- "expose-loader"
- "uglifyjs-webpack-plugin@~0.4.6"

PR #543, opened 2017-07-19
Adding versions to the NPM packages

apaganobeleno:
Given we've seen issues where some webpack/node package change and newly created buffalo app stop working we need to ensure our versions are stablished when generating the package.json file.

This PR sets these versions leaving room for minor changes in the packages.


apaganobeleno:
Related with #503

PR #548, opened 2017-07-20
Fix panic on path permissions issue

ryanfaerman:
If a directory in the path has the incorrect permissions, the `info` variable will be `nil`, causing a panic.

I figured it would be better to just skip that directory.


markbates:
Thanks. Can you update this to work with the `development` branch? Thanks.

markbates:
I manually add this. Thanks.

PR #553, opened 2017-07-23
Tab completions

hasit:
1. Add tab completion file for fish shell in `/completions/fish/buffalo.fish`
2. Add grift for installing buffalo.fish in `/grifts/completions.go`
3. Add instructions for installing buffalo.fish in `/completions/fish/README.md`

### DIFF
--- a/completions/fish/README.md
+++ b/completions/fish/README.md
@@ -1,0 +1,53 @@
+# Install
+
+## Manually
+
+1. See where your `fish` configuration stores completion scripts
+
+```
+~ echo $fish_complete_path
+/Users/hasitmistry/.config/fish/completions /usr/local/etc/fish/completions /usr/local/share/fish/completions /Users/hasitmistry/.local/share/fish/generated_completions
+```
+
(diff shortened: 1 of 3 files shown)


markbates:
While I love the spirit of this PR, I'm not going to merge it. Let me explain why.

The first reason is, this isn't the optimal way to achieve this. This would've been better as a sub-command of `generate`. For example, `$buffalo generate fish`. Something like that. Having people `cd` into the Buffalo project folder to run a task isn't the best experience.

The second reason, the more I think about this, the more I think this would be better suited as a plugin for buffalo, not part of the Buffalo project itself. Plugins are meant for exactly this type of extending of the project.

Documentation on how to write plugins can be found at [http://gobuffalo.io/docs/plugins](http://gobuffalo.io/docs/plugins).

hasit:
@markbates 

I do see why it would make more sense to add it as a plugin. As a plugin, we would be able to support completion for:

1. Commands/flags when they are added/modified in the future 
2. Plugins added by the users

PR #559, opened 2017-07-25
Removes extra ? when the route helper doesn't have extra args

apaganobeleno:


apaganobeleno:
thanks to @SantoDE for pointing this one out.

PR #565, opened 2017-07-28
fixed unwanted trailing query-sign and repeated reassignment

sio4:
Hi, recently new function `addExtraParamsTo` was added and it very cool! but currently, it makes extra query sign(`?`) in end of all generated URL even if there is no extra options. I fixed that with simple length checking of pendingParams's key list.

Additionally, since it called from template, query escaping is somewhat annoying. I just added query escaping for string type into same commit.

Last part of the `diff` is just speed up in mind :-) There is only one 'the first' element on the array, so reassign to the first is more cheaper than reassign to the others.

Please check and consider the pull request.

### DIFF
--- a/route.go
+++ b/route.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"net/url"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


apaganobeleno:
@sio4 thanks for the PR, this one was already fixed by #559 and #560, which were already merged into development.

sio4:
Oh! I didn't know although I've turned on the notification from here. OK, I will wait next release :-)

PR #569, opened 2017-07-29
Add Context.Cookies

### LINKED ISSUE #391: Create a new context.Cookie wrapper for gorilla Cookiestore
There is a context variable for `gorilla` package's `SessionStore`. We need to add similar `CookieStore` handler to the context variable or app as the whole.

bigblind:
This gives easy access to request cookies, and makes it easy to set cookies on the response.

Closes #391

### DIFF
--- a/context.go
+++ b/context.go
@@ -18,6 +18,7 @@ type Context interface {
 	Response() http.ResponseWriter
 	Request() *http.Request
 	Session() *Session
+	Cookies() *Cookies
 	Params() ParamValues
 	Param(string) string
 	Set(string, interface{})
--- a/cookies.go
+++ b/cookies.go
@@ -1,0 +1,59 @@
+package buffalo
+
+import (
+	"net/http"
+	"time"
+)
+
+// Cookies allows you to easily get cookies from the request, and set cookies on the response.
+type Cookies struct {
+	req *http.Request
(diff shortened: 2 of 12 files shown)


markbates:
This looks great! Should we have a function that let's people set an `http.Cookie`? I love the helper methods for the common use case, but also being able to set a complex cookie the user defines would be great.

bigblind:
I didn't add this, because I thought these people would just use
http.SetCookie directly. The whole reason for this wrapper to exist, is so
people don't have to manually construct cookies, or even import the http
package. If I were to add this method, it'd simply call http.SetCookie with
the context's responseWriter, and the passed-in cookie. The only advantage
of calling this wrapper rather than the original, is that you don't have to
pass in the response, but I don't think that's a huge benefit.

But ultimately, it's simple to implement, and doesn't add any complexity,
so I can implement it if you want.

markbates:
@bigblind good point.

markbates:
@bigblind It looks like your branch is in some crazy weird state. It has huge conflicts with `development` and says 25 files have changed and such. Can you figure out what's going on there so I can merge? Thanks.

bigblind:
@markbates, I'm sorry, I messed up a bit. should be fixed now.

Issue #579, opened 2017-08-04
Verbose log of workers


sio4:
Hi,
Today I tried to use the simple worker. and I found the log level for "Performing" (worker/simple.go#L75) and "Completed" (worker/simple.go#L89) is Info. I think it can be Debug level because this logs need to be suppressed when I put this app in production.

markbates:
A PR would be welcome for this.

sio4:
OK, I will do this tonight.

sio4:
Hi, @markbates I did branched from current development branch but it seems there is some missing commit not applied to development but exists on master. Can I do cherry pick as below before I start? This commit 5aa1cff287e97fd4f8e2169abbe3c6bd4bba9220 is for the path of `logrus`.

`git cherry-pick 5aa1cff287e97fd4f8e2169abbe3c6bd4bba9220`

sio4:
Oops! I found the code on 'development' branch is also slightly outdated. Can I make PR against master for this? or Can I merge your works d7390f29d5c1626b7e7879433301fae3cad992ff and 093e7f3b6c626e85511e2f13335623075ad0f819 before I start?

sio4:
I removed two of my own comments because it's my fault.
I just made simple changes on current development branch and requested a PR.

sio4:
Thanks for merging it quickly. I hope it will get on the master soon.
Thanks.

markbates:
Thank you! The next release is coming up soon. So it won't be a long wait. :)

PR #581, opened 2017-08-06
changed log level for perform method of simple worker.

sio4:
Hello, this PR is for #579.

- make log as debug not info.
- remove trailing newline.


sio4:
Second commit contains a patch related to auto-generated actions/resources.go by buffalo generator.
The template has two typos:

- default value for `per_page` on pop is 20, not 2.
- auto generated field name for creation time is `created_at` not `create_at`.

sio4:
Thanks

PR #582, opened 2017-08-08
Skip the windows directory when looking for plugins

ryanfaerman:
The buffalo command can seemingly hang when looking for plugins on some Windows systems. The `%PATH%` could have multiple entries containing `C:\windows` causing multiple file walks of the same directories.

Filewalking that path is fairly expensive, as it has many files and folders, most of which are Windows internals.

Adding this change dropped the startup time from ~30s to less than a second.


markbates:
Looks like code climate is failing.

ryanfaerman:
@markbates The line that Code Climate is failing on, isn't one that I changed. I'll run `gofmt` nevertheless.

PR #588, opened 2017-08-15
When fetching packages with go get, remove redundant go install

bradleyfalzon:
"go get" fetches and installs the specified package according to the
documentation:

    Get downloads the packages named by the import paths, along with their
    dependencies. It then installs the named packages, like 'go install'.

Testing this myself on go1.8.3 and go1.9rc2 having an older version of
buffalo binary install, then running:

    go get -u github.com/gobuffalo/buffalo/...

Shows the binary is also updated successfully when checking the version.

Relates to documentation update: https://github.com/gobuffalo/gobuffalo/pull/57

Further, this appears to have first started with https://github.com/gobuffalo/buffalo/commit/1bcb98153c42d8c0c2c1243c02c5338ca2c55886#diff-00e5b9d3695f2772e4d4422b4fee4787R92 where there's a `glide get` of `github.com/markbates/refresh` followed by `go install` of the same package. Could this have initially been a `glide` vs `go` behaviour difference that continued through the various refactors? But it seems safe to remove the additional `go install` now.

I have tested this on a basic `buffalo new <name>`, but I haven't extensively tested this change in regards to the soda related changes or Dockerfile.

### DIFF
--- a/buffalo/cmd/new.go
+++ b/buffalo/cmd/new.go
@@ -180,7 +180,7 @@ func init() {
 	RootCmd.AddCommand(newCmd)
 	newCmd.Flags().BoolVar(&app.API, "api", false, "skip all front-end code and configure for an API server")
 	newCmd.Flags().BoolVarP(&app.Force, "force", "f", false, "delete and remake if the app already exists")
-	newCmd.Flags().BoolVarP(&app.Verbose, "verbose", "v", false, "verbosely print out the go get/install commands")
+	newCmd.Flags().BoolVarP(&app.Verbose, "verbose", "v", false, "verbosely print out the go get commands")
 	newCmd.Flags().BoolVar(&app.SkipPop, "skip-pop", false, "skips adding pop/soda to your app")
 	newCmd.Flags().BoolVar(&app.WithDep, "with-dep", false, "adds github.com/golang/dep to your app")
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -17,8 +17,6 @@ ADD . .
 
 RUN go get -v -t $(go list ./... | grep -v /vendor/)
 
-RUN go install -v ./buffalo
-
```
markbates: This is needed. We run our tests using this dockerfile so we want to make sure we are running the tests using the code that was just committed.

bradleyfalzon: Sure, but are you certain it's needed? The go get should be handling this. But I can add it back, no issue.

markbates:
Definitely needed. It's compiling the local source.

Dockerfile
bradleyfalzon: > Definitely needed. It's compiling the local source.

OK, but I've tested by adding a panic to new generator and the `buffalo new` commands started to panic. The go get should still install the local version.

But I'll add it back now.

bradleyfalzon:
I've changed the base branch and removed the Dockerfile modifications, ready for a second review.

markbates:
Thanks!

PR #589, opened 2017-08-16
Added Body print capabilities to ParamLogger if the request is application/json

dexafree:
While using Buffalo for developing an API, I found myself in the need of inspecting the JSON POST requests in order to detect some anomalies, and I was surprised that the ParamLogger printed the URL, the FORM and the QueryString params, but didn't print the body of the POST requests that weren't FormUrlEncoded.

This Pull Request adds a check to detect if the request has a Content-Type of "application/json", and if it has, it dumps the JSON. Otherwise, it's assumed that it's a FormUrlEncoded request and prints the Form as it did before.

### DIFF
--- a/middleware/param_logger.go
+++ b/middleware/param_logger.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"encoding/json"
+	"io/ioutil"
 
 	"github.com/gobuffalo/buffalo"
 )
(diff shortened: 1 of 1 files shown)


dexafree:
Closing it as it might cause `req.Body` to throw an error if needed after logging.

Sorry for not noticing it previously.

PR #611, opened 2017-08-30
Use Accept instead of Content-Type

l4u:


markbates:
It appears that the tests are failing for this.

PR #615, opened 2017-09-03
Fix 'formular' typo

slomek:
Not a 100% sure since I'm not an English native speaker, but Cambridge Dictionary [agrees](http://dictionary.cambridge.org/spellcheck/english/?q=formular) that there's not such a word...


markbates:
Thanks!

PR #619, opened 2017-09-05
fix for gorilla context memory leak

nutbunnies:
After using the `Null` session I noticed that memory usage dropped considerably.  This lead to looking into memory usage of a fresh buffalo app (`buffalo new --api --skip-pop memtest`) and found that memory usage grows with each request and it's related to the gorilla context / session code.  They call out several memory issues related to the shallow copy of the request when mixing gorilla context with go context.  This PR moves the gorilla `context.Clear` to the buffalo handler before it creates it's own context.

http://www.gorillatoolkit.org/pkg/sessions
https://github.com/gorilla/sessions/pull/110

Screenshot is from a heap after 200k requests to the default home controller -- after the change these nodes were no longer present.
screen shot 2017-09-05 at 11 56 52 am

NewRelic goagent shows the memory growth during load test:
screen shot 2017-09-05 at 8 17 13 am

After the fix the load test shows stable memory usage:
screen shot 2017-09-05 at 12 18 40 pm

### DIFF
--- a/app.go
+++ b/app.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"syscall"
 
-	gcontext "github.com/gorilla/context"
 	"github.com/gorilla/mux"
 	"github.com/markbates/refresh/refresh/web"
 	"github.com/markbates/sigtx"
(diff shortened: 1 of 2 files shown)


markbates:
Now that's how you write a PR!! Awesome! Thanks.

PR #621, opened 2017-09-07
add new option to not override url names

SantoDE:
Hey @markbates,

that is some kind of an idea to make it possible to reuse the same url names even though you may be running the root app with a prefix. Not sure if it's good, but it works out at least :)

Looking forward to your feedback!

### DIFF
--- a/options.go
+++ b/options.go
@@ -52,6 +52,7 @@ type Options struct {
 	Context context.Context
 	cancel  context.CancelFunc
 	Prefix  string
+	URLNameOverride bool
 }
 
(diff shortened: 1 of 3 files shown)


markbates:
I'm not a fan of this approach, but I currently don't have a good alternate approach. I think part of the problem is I don't like the idea of another option, particularly one that is a `bool`. My current thinking is that the function that defines the route name should maybe a `var` so that those who need too can set their own function to define how those names are set. But even that, I'm not a fan of. Give me some time to think about it.

PR #622, opened 2017-09-07
add posibility to recirect by path name

SantoDE:
Hey @markbates ,

and another proposal :) That one adds the functionality to redirect by a path name. WDYT?


markbates:
Replaced by #1020

Issue #623, opened 2017-09-07
Build on Windows produces incorrect extension


hut8:
I hate to be _that guy that uses Windows_ , but here I am 😄 

After creating a new project, if I run `buffalo b` on Windows, I get bin/my-project instead of bin/my-project.exe 

Looks like this line is at fault: https://github.com/gobuffalo/buffalo/blob/master/buffalo/cmd/build.go#L422

It's a small bug; I believe that replacing `os.Getenv("GOOS")` with `runtime.GOOS` should work.

markbates:
Any chance I could convince you to submit a PR for it? :)

-----------
Mark Bates

hut8:
Indeed. Will submit shortly.

PR #624, opened 2017-09-07
Check GOOS from runtime rather than env var

hut8:
Resolves https://github.com/gobuffalo/buffalo/issues/623


markbates:
Thanks Liam! Feel free and keep those Windows PRs coming! :)

PR #635, opened 2017-09-13
Feature assets fingerprint

paganotoni:
Adds assets fingerprinting back.


markbates:
let's try this again. :)

PR #636, opened 2017-09-13
deployed binary cannot pass in argument to task fixes #625

### LINKED ISSUE #625: deployed binary cannot pass in argument to task
buffalo v0.9.3

i just confirmed that the issue isnt about heroku.
i tried running this on my docker:
and my build app has been using heroku name(just naming)
./bin/app task user:create xxx

my task user:create is executed, but all arguments is missing
xxx wasn't passed in

markbates:


paganotoni:
Great one Mark! thanks for it

PR #642, opened 2017-09-18
refactored app.Start to gracefully shutdown

nutbunnies:
Noticed that requests aren't allowed to finish durning shutdown even though `server.Shutdown` was called.  This was due to `ListenAndServe` returning once the shutdown starts, this causes the application to terminate.

This basically flips the functionality that was in a goroutine in `app.Start` adding a channel to get errors from `ListenAndServe` and making the timeout for shutdown configurable.  This only applies to http requests finishing, if there are no open connections but the worker is active the application will still exit immediately.

Test verifies correct shutdown behavior by actually starting the application -- guessing there is a much better and less brittle way to verify.

### DIFF
--- a/app.go
+++ b/app.go
@@ -3,13 +3,16 @@ package buffalo
 import (
 	"context"
 	"fmt"
+	"net"
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
 	"runtime"
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/markbates/refresh/refresh/web"
@@ -30,6 +33,8 @@ type App struct {
 	routes        RouteList
 	root          *App
 	children      []*App
+	server        http.Server
+	closed        bool
 }
 
 // Start the application at the specified address/port and listen for OS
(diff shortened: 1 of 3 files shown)


options.go
```diff
@@ -115,5 +120,6 @@ func optionsWithDefaults(opts Options) Options {
 	}
 	opts.SessionName = defaults.String(opts.SessionName, "_buffalo_session")
 	opts.Host = defaults.String(opts.Host, envy.Get("HOST", fmt.Sprintf("http://127.0.0.1:%s", envy.Get("PORT", "3000"))))
+	opts.ShutDownTimeout = 30 * time.Second
```
markbates: This will overwrite any setting a user has set. you should check whether it has been set already or not.

app_test.go
```diff
@@ -1,5 +1,72 @@
...
+
+	go func() {
+		time.Sleep(1 * time.Second)
+		_, err := http.Get("http://localhost:17171/slow")
+		if err != nil {
+			t.Error("unexpected error requesting slow endpoint", err)
+		}
+	}()
+
+	app.Start(":17171")
```
markbates: Not a fan of the hard coded port here. Perhaps we can do what they do with `httptest.NewServer` https://golang.org/pkg/net/http/httptest/#NewServer

app_test.go
```diff
@@ -1,5 +1,72 @@
...
+
+	app := New(NewOptions())
+	app.GET("/slow", func(c Context) error {
+		time.Sleep(4 * time.Second)
+		requestFinished = true
+		return nil
+	})
+
+	go func() {
+		time.Sleep(2 * time.Second)
```
markbates: Can we reduce these sleep times? They really slow down the tests.

app_test.go
```diff
@@ -1,5 +1,72 @@
...
+		app.Stop(nil)
+	}()
+
+	go func() {
+		time.Sleep(1 * time.Second)
+		_, err := http.Get("http://localhost:17171/slow")
+		if err != nil {
```
markbates: Can you please use the `require` assertion library, like the rest of the tests? We like to keep things consistent.

app.go
```diff
@@ -39,38 +39,20 @@ func (a *App) Start(addr string) error {
 	if !strings.Contains(addr, ":") {
 		addr = fmt.Sprintf(":%s", addr)
 	}
-	fmt.Printf("Starting application at %s\n", addr)
+	a.Logger.WithField("interface", addr).Info("Starting application")
```
markbates: Is there a reason you changed all of the logging?

markbates: In particular, I'm curious about why the capitalization is different.

markbates:
It appears as though this introduced a race condition and the tests are failing.

app.go
nutbunnies: Ended up adding a ton of logging to diagnose and the habit of only using the configured logger kicked in.  Have a personal nit when unstructured or inconsistent logging structures messages are emitted. :)

Capitalization changes are just from retyping -- not on purpose.

It's unrelated to issue -- I'll revert

app_test.go
nutbunnies: will do -- `require` readability is much better

options.go
nutbunnies: Doh.... I'll open a PR for the `going` helpers to keep the default checks together

app_test.go
nutbunnies: Me either.  From a quick dive into NewServer and google looks like using `:0` to get a dynamic port is the way to go but will require creating the Listener to pull it later.

https://golang.org/src/net/http/httptest/server.go?s=2300:2344#L55
https://stackoverflow.com/questions/31632506/how-to-get-address-for-http-listenandserve
https://stackoverflow.com/questions/43424787/how-to-use-next-available-port-in-http-listenandserve

nutbunnies:
The data race looks to be from the worker process mutating the share context passed on startup.  Disabling the worker and/or adding locks around start and stop in the app and simple worker appear to fix it.  Not my favorite.

I think this is going to be bigger than just moving the code around in the app.Start function.    Based on the article you mentioned in slack (https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation) - What do you think about removing the context all together from the app.start function?  Removing it would create a breaking change to the `Worker` interface.

PR #649, opened 2017-09-19
Fix API mode import errors

### LINKED ISSUE #648: Import error generating new app as API only
I'm seeing issues similar to #492 using v0.9.4. 
`actions/render.go:5:2: imported and not used: "github.com/gobuffalo/packr"`

It appears to be related to newapp/templates/actions/render.go.tmpl importing packr in API mode, where the packr use is left out when in API mode.

I've got a fix I'll put a PR in for in just a minute to get this sorted out. Hope that works!

tsmith1024:
Fixes #648

### DIFF
--- a/generators/newapp/templates/actions/render.go.tmpl
+++ b/generators/newapp/templates/actions/render.go.tmpl
@@ -1,7 +1,9 @@
 package actions
 
 import (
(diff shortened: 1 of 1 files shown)


markbates:
Thanks!

PR #655, opened 2017-09-23
fix: remove mixed tabs and spaces from .travis.yml template

DanielHeckrath:
Travis CI builds for new projects fail because of tabs in the generated .travis.yml file


app.go
```diff
@@ -101,7 +100,6 @@ func (a *App) Stop(err error) error {
 }
 
 func (a *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	defer gcontext.Clear(r)
```
markbates: what is this change for?

handler.go
```diff
@@ -63,6 +64,7 @@ func (a *App) newContext(info RouteInfo, res http.ResponseWriter, req *http.Requ
 }
 
 func (info RouteInfo) ServeHTTP(res http.ResponseWriter, req *http.Request) {
+	defer gcontext.Clear(req)
```
markbates: and this change?

PR #658, opened 2017-09-25
Added go-i18 to the SHOULDERS

wideawake:


markbates:
@wideawake this is an autogenerated file, so this change will be overwritten when the `grift shoulders` task is next run.

To make the change permanent you'll want to add it here: https://github.com/gobuffalo/buffalo/blob/master/grifts/shoulders.go#L26

PR #669, opened 2017-09-30
Fix links in SHOULDERS.md

jorinvo:
Links pointing to sub-packages were broken.
Not sure if this file is generated, but I fixed the links manually for now.


markbates:
Thanks. It is generated and we fixed the generator so when we run it for the next release it should be good.

PR #674, opened 2017-10-10
speed up integration tests

markbates:


markbates:
Offered no speed improvements.

Issue #683, opened 2017-10-11
assets uglifier for production only?


u007:
hi,

new assets fingerprint and uglifier is awesome 👍 
managed to reduced js from 4mb to 1.6mb.

but there are a concern of performance for development.
the speed to recompile all assets took very long.

is it possible to have webpack command for different environment?
development, staging, production

maybe in webpack.yml (GO_ENV):
```
development: webpack --env.NODE_ENV=$GO_ENV -c ./webpack.config.js -watch
build: webpack --env.NODE_ENV=$GO_ENV -c ./webpack.config.js
```

with this, it will be awesome that we only run build to do uglifier and clean up
what do you think?

here are example:
https://webpack.js.org/guides/environment-variables/

thank you

u007:
my build now increase to 200+mb

u007:
i need a urgent fix for this.
thinking to find if webpack.$GO_ENV.config.js exists, otherwise use webpack.config.js to build

is this good?

markbates:
@u007 https://github.com/gobuffalo/buffalo/pull/687 - I think that's a better solution.

PR #686, opened 2017-10-12
webpack 3 and webpack.production.config.js support

u007:
production webpack with: cleanup and uglifier
default webpack does not contains uglifier

as requested in https://github.com/gobuffalo/buffalo/issues/683

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,19 @@ Just like running your application, Buffalo doesn't stop you from using the stan
 $ buffalo test
 ```
 
+## To build for deployment
+
+to compile binary:
+```text
+$ mkdir bin
+$ GO_ENV=production GOOS=linux buffalo build -o bin/app
+```
+
+to cross compile for linux
+```text
+$ GOOS=linux GO_ENV=production GOOS=linux buffalo build -o bin/app
+```
+
 ## Shoulders of Giants
 
 Buffalo would not be possible if not for all of the great projects it depends on. Please see [SHOULDERS.md](SHOULDERS.md) to see a list of them.
--- a/README.md
+++ b/README.md
@@ -67,12 +67,12 @@ $ buffalo test
 to compile binary:
 ```text
 $ mkdir bin
-$ GO_ENV=production GOOS=linux buffalo build -o bin/app
+$ GO_ENV=production buffalo build -o bin/app
 ```
 
 to cross compile for linux
 ```text
-$ GOOS=linux GO_ENV=production GOOS=linux buffalo build -o bin/app
+$ GOOS=linux GO_ENV=production buffalo build -o bin/app
 ```
 
 ## Shoulders of Giants
--- a/buffalo/cmd/build.go
+++ b/buffalo/cmd/build.go
@@ -61,11 +61,25 @@ func (b *builder) execQuiet(name string, args ...string) error {
 }
 
 func (b *builder) buildWebpack() error {
-	_, err := os.Stat("webpack.config.js")
+	envName := envy.Get("GO_ENV", "development")
+	configFile := "webpack.config.js"
+
+	fmt.Println("build: environment: ", envName)
+	if envName != "" {
+		testConfigName := "webpack." + envName + ".config.js"
+		_, err := os.Stat(testConfigName)
+		if err == nil {
+			// build webpack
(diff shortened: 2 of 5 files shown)


u007:
tested with new project, and build for production
build binary tested working

markbates:
I'm not merging this PR for a few reasons. The first is you are doing too many things in this PR; you introduce new web pack files, as well as a new version of web pack. These should be different PRs. Also, this PR is against `master` all PRs need to be against the `development` branch, especially since yesterday we merged in a new build system, which this touches.

The last reason is I'm not particularly a fan of having all these random new files, plus, there are a lot of places in the code that are looking `webpack.config.js` we need to make sure we are fixing all of those places to look for different versions based on environment.

I'm not a webpack expert, not by any means, but I wonder if there's a way to turn on/off plugins within web pack using a flag. I'd be all about that. If we could pass a `production` flag to web pack during build and it would do the right thing, that would be the solution I would love. 

If that isn't possible, then we need to look at alternate solutions.

majorcode:
In another project I control webpack build mode with an environment variable. You can access the environment in `webpack.config.js` via `process.env.<var-name>`. Below is an example from my project. I set NODE_ENV on the webpack command line. @markbates is this the way to turn on/off plugins you were looking for? Maybe you already set some environment variable when running webpack in `buffalo dev` mode that can be used in `webpack.config.js`.

```javascript

var webpack = require('webpack');
var path = require('path');

var sourcePath = path.join(__dirname, './assets');
var staticsPath = path.join(__dirname, './public/assets');

module.exports = function () {
  var nodeEnv = process.env.NODE_ENV;
  var isProd = nodeEnv === 'production';

  var plugins = [
  ];

  if (isProd) {
    let CompressionPlugin = require('compression-webpack-plugin');

    plugins.push(
      new webpack.LoaderOptionsPlugin({
        minimize: true,
        debug: false
      }),
      new webpack.optimize.UglifyJsPlugin({
        sourceMap: true,
        compress: {
          warnings: false,
          screw_ie8: true,
          conditionals: true,
          unused: true,
          comparisons: true,
          sequences: true,
          dead_code: true,
          evaluate: true,
          if_return: true,
          join_vars: true,
        },
        output: {
          comments: false,
        },
      }),
      new CompressionPlugin({
        asset: "[path].gz[query]",
        algorithm: "gzip",
        test: /\.js$|\.css$|\.html$/,
        threshold: 10240,
        minRatio: 0.8
      })
    );
  }

  return {
    devtool: isProd ? 'source-map' : 'eval',

    context: sourcePath,

    entry: {
      "common": "./common",
      "reader": "./reader"
    },

    output: {
      path: staticsPath,
      filename: '[name].js',
    },

    module: {
      rules: [
        {
          test: /\.scss$/,
          use: [
            'style-loader',
            'css-loader',
            'sass-loader'
          ]
        },
        {
          test: /\.css$/,
          use: [
            'style-loader',
            'css-loader'
          ]
        },
        {
          test: /\.js$/,
          exclude: /node_modules/,
          use: [
            'babel-loader'
          ],
        },
        {
          test: /\.gif$/,
          use: [
            "url-loader"
          ]
        }
      ],
    },

    resolve: {
      extensions: ['.webpack-loader.js', '.web-loader.js', '.loader.js', '.js', '.jsx'],
      modules: [
        path.resolve(__dirname, 'node_modules'),
        sourcePath
      ]
    },

    plugins,

    performance: isProd && {
      maxAssetSize: 700000,
      maxEntrypointSize: 700000,
      hints: 'warning',
    },

    stats: {
      colors: {
        green: '\u001b[32m',
      }
    }
  };
};


```

My webpack foo is not too strong. But I think this points in the right direction.

markbates:
@majorcode I basically came up with the same solution here: https://github.com/gobuffalo/buffalo/pull/687

u007:
i rather prefer a separate file.
having one file that control everything is messy

iam going to use my own repository

majorcode:
@u007 there are up sides to having a single file. You can see all of the config in one place. Also, since the production switch only adds a few things, none of the common configuration is duplicated. Just a thought.

u007:
@majorcode okay noted :)

btw, the latest update need to remain the magnifest config for development

PR #687, opened 2017-10-12
assets uglifier for production only? closes #683

### LINKED ISSUE #683: assets uglifier for production only?

markbates:


buffalo/cmd/build/assets.go
```diff
@@ -2,12 +2,14 @@ package build
...
 import (
 	"github.com/gobuffalo/buffalo/generators/assets/webpack"
+	"github.com/gobuffalo/envy"
 	pack "github.com/gobuffalo/packr/builder"
 	"github.com/pkg/errors"
 )
 
 func (b *Builder) buildAssets() error {
 	if b.WithWebpack {
+		envy.Set("NODE_ENV", "production")
```
u007: hi Markbates, 
would be nice if it detects GO_ENV and set it  with default to "production"
and set it into NODE_ENV

generators/assets/webpack/templates/webpack.config.js.tmpl
```diff
@@ -44,48 +104,6 @@ module.exports = {
         screw_ie8: true
       },
       comments: false
-    }),
-    new ManifestPlugin({
```
u007: manifest is needed for application.[hash].js to work

markbates: It’s still there. Look at the whole file.

u007: ohh ya

PR #691, opened 2017-10-16
wow. ok. that was a lot of work, but i think it's going to be totally worth it.

markbates:


markbates:
Clean up commands so they can be called programmatically. fixes #692

Issue #700, opened 2017-10-20
[feature] build only binary without assets compilation


u007:
hi,

is it supposed to have buffalo build -e working this way?
ive tried, and realize it still call webpack compilation

is there any other flag to just compile binary for non assets handling?
thanks

stanislas-m:
The `-e` option for `buffalo build` extracts the assets to a zip archive, instead of linking it in the binary. These assets still needs to be build, that's why the webpack compilation is still called.

markbates:
This would be a pretty easy PR for anyone looking to do it. All it would be is added a new flag and turning off that part of the build system. The build system has been rewritten in `development` and is a lot easier to work with now.

paganotoni:
I would love to work on this one

PR #702, opened 2017-10-20
Allows to pass -k to disable assets compilation

paganotoni:
This is related to #700

### DIFF
--- a/buffalo/cmd/build/builder.go
+++ b/buffalo/cmd/build/builder.go
@@ -38,12 +38,6 @@ func New(ctx context.Context, opts Options) *Builder {
 		b.buildBin,
 	}
 
-	if b.Options.SkipAssets {
-		logrus.Debug("skipping assets compilation step")
-		b.steps = append(b.steps[:6], b.steps[7:]...)
-		return b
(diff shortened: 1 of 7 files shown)


buffalo/cmd/build/builder.go
```diff
@@ -38,6 +38,12 @@ func New(ctx context.Context, opts Options) *Builder {
 		b.buildBin,
 	}
 
+	if b.Options.SkipAssets {
```
markbates: This won't work correctly. If you do this here then the templates, and anything else that uses `packr` won't run. The correct thing to do is to put the logic in the `buildAssets` method.

paganotoni: I see, ok, i changed the code to do the check inside the buildAssets function, LMK if it makes sense.

markbates:
I think the logic looks good. My advice would be to keep the flag as is, but change the name of the option to be `WithAssets` that way the logic reads easier. Does that make sense?

paganotoni:
Sounds good, going for it, thanks for all the patience and guidance on this one Mark

PR #704, opened 2017-10-22
App binds to 127.0.0.1:3000 by default

adamlamar:
In `options.go`, the `Options.Host` field states the default bind location is `127.0.0.1:3000`. However, this value doesn't seem to be used anywhere in buffalo.

I noticed that I was presented with Mac's firewall prompt every time my buffalo app started. This prompt isn't presented when a program binds to `127.0.0.1` only, so the app must be binding to `0.0.0.0`.

firewall

Rather than use the bind address/port provided to `app.Start()`, this PR adjusts `app.Start()` to use the `Options.Host` value. This should provide the intended behavior from the `Options.Host` description.

It also removes the argument to `app.Start()`, which would be an issue for app upgrades where `main.go` was generated to provide the `addr` argument. 

After building a test project with this change, the app binds to `127.0.0.1:3000` by default. `PORT` can still be overridden as before, and the Mac's firewall prompt no longer pops on app start.

### DIFF
--- a/app.go
+++ b/app.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


markbates:
Hi, thanks for the PR. A couple of things, first, all PR's need to be against the `development` branch, as noted in the CONTRIBUTING.md file. I've changed the branch, but now there are conflicts that need resolution. 

Second the tests on Travis are failing, so those need to be addressed. You can run the full set of integration tests locally, with `docker build .`.

Finally, the `Host` option is meant to be the "public" host, so things like `https://foo.bar`, not the host to bind to. By default, Go, does bind to `0.0.0.0` if there is not host present. I'm up for setting the ability to change that, but I think we need a different option for that, perhaps something like `ServerHost` or `BindHost`, something along those lines. Thoughts?

markbates:
I like Addr. That’s good. Let’s use that.

By public url I meant the actual address of their site.

-----------
Mark Bates

adamlamar:
@markbates I rebased this branch against `development`, force-pushed to my fork's branch and addressed the merge conflicts. Also used `Addr` instead of `Host`. Mind taking another look?

markbates:
I created a new PR for this #719 because I wanted to create some deprecations around the `Start` function in particular, since it was a breaking change. Hope you don't mind.

PR #706, opened 2017-10-23
Cleaning up stale webpack output.

barrongineer:
Asset fingerprinting was causing `public/assets` to fill up with stale files when running `buffalo dev` and making js/css edits. Added a webpack plugin and config to clean out the js and css output before generating new files.


markbates:
love it! thanks!

PR #718, opened 2017-10-25
Add support for Bazaar VCS when creating a new app

stanislas-m:
* Add support for [Bazaar](http://bazaar.canonical.com/en/) VCS
* Add a new --vcs flag, with defaults to git (non-regression)
* Add none & bzr options to --vcs choices


generators/newapp/new.go
```diff
@@ -110,10 +110,20 @@ func (a Generator) Run(root string, data makr.Data) error {
 		},
 	})
 
-	if _, err := exec.LookPath("git"); err == nil {
+	if a.VCS == "git" {
```
markbates: I think we can replace this whole if block.

First, let's return early if the `a.VCS` is `""` or `"none"`.

Then we can replace this big `if` block with something like:

```go
	if a.VCS == "" || a.VCS == "none" {
		return nil
	}

	if _, err := exec.LookPath(a.VCS); err == nil {
		g.Add(makr.NewCommand(exec.Command(a.VCS, "init")))
		g.Add(makr.NewCommand(exec.Command(a.VCS, "add", ".")))
		g.Add(makr.NewCommand(exec.Command(a.VCS, "commit", "-m", "Initial Commit")))
	}
```

that works for now since, according to this PR, both `git` and `bzr` have the same CLI API.

stanislas-m: For the first point, we can't return early from here, we still have to execute the generator.

I agree the big `if` block can be factorized, but I'm not fan letting all strings working. The new commit  filters the supported options, and factorize the code.

stanislas-m:
@markbates Does this new version seems OK to you?

markbates:
👍

PR #729, opened 2017-11-01
Update README.md - Fix for installation procedure

rgerardi:
I have tested this installation procedure in 3 different operating system and it is the most reliable way. Current installation method causes issues with some dependencies.
This change also synchronizes the procedure with `gobuffalo.io` to avoid confusion.


paganotoni:
@markbates adding you as reviewer of this one given to me is very sensible :)

markbates:
The following:

```bash
$ go get -u -v github.com/gobuffalo/buffalo/buffalo
```

Should pull down the code, as well as, install the binary.

rgerardi:
Hi Mark
By running the following command alone, there are missed dependencies.
```
$ go get -u -v github.com/gobuffalo/buffalo/buffalo
```
I have validated it installing `buffalo` in 3 different boxes that never had it installed before (2 Linux and 1 MacOS).
For example, the package `github.com/gobuffalo/x/httpx` was missing.

Installing using the sequence I propose was the most reliable way to get `buffalo` working in all 3 boxes.

markbates:
Awesome. Sounds good to me. :) Thanks for the investigation on this. That’s really helpful. Thanks!

PR #735, opened 2017-11-04
Fixing underscored multi word resource generation

### LINKED ISSUE #734: Generating a resource composed by 2 words underscored
When generating a resource with 2 words where the words are separated with underscore buffalo generates views, model, and all necessary files, however, it uses underscored version in some places and camelcase in other, causing the resource not to work correctly.

````
buffalo g r person_event type description person_id:uuid
````

Uses `person_event` on the actions but uses `personEvent` on the view, causing some errors when running the app.

It should use `person_event` and `personEvent`.

paganotoni:
Makes actions set the correct variable name when generating resources and the user has passed underscored name on the resource name.

This closes #734


paganotoni:
@markbates this one is ready to be reviewed

PR #761, opened 2017-11-17
no longer need extensions for templates in actions or partials for HTML, JS, and Markdown

markbates:
With this change you no longer are required to add extensions if the template has the same content type as the template it's being called from, or the render engine calling it. This only works with HTML, JavaScript, and Markdown.

```erb
<body>
  <%= partial("nav") %>

  <div class="container" id="main">
    <%= partial("flash") %>
    <%= yield %>
  </div>

  <%= partial("footer") %>

</body>
```

```go
func HomeHandler(c buffalo.Context) error {
	return c.Render(200, r.HTML("index"))
}
```

### DIFF
--- a/render/html_test.go
+++ b/render/html_test.go
@@ -2,8 +2,8 @@ package render_test
 
 import (
 	"bytes"
-	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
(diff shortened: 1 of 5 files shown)


bketelsen:
nice

PR #763, opened 2017-11-18
Simplify i18n

stanislas-m:
- Use the `t` helper for every case (no more `tp`).
- Load languages & translator function in the middleware, so the Translate function can be used inline; even in Go code.
- Get closer to the i18n function interface, by forwarding variadic arg (plush already fixed for that).
- Use the user data as context values, not Buffalo context (not precise enough in a loop, for instance).

This introduces a breaking change, because the `tp` helper is removed, and the i18n helpers return only a string now.


markbates:
I think these tests are failing because the docker image doesn't have the latest version of Plush on it. I just kicked off a rebuild of the image on docker hub. Once that's done you can retry the tests.

https://hub.docker.com/r/gobuffalo/buffalo/builds/

markbates:
Other than the tests, things look good, so I'll happily merge once the tests are passing.

stanislas-m:
@markbates Seems good now! :)

Issue #751, opened 2017-11-12
Error: unknown flag: --use-model


kberov:
Hello,
I already have a generated model like this:
```
buffalo db g model -s users id:uuid login_name:string login_password:string \
    first_name:string last_name:string email:string description:string \
    created_by:uuid changed_by:uuid start_date:int stop_date:int
```
Now I want to use it to create the rest of the resource.
`buffalo help g r` has an example:
```
$ buffalo g resource users --use-model users
Generates:

- actions/users.go
- actions/users_test.go
```
I run `buffalo g r users --use-model users` and I get the help message with the above mentioned error at the bottom.
Buffalo version is: v0.10.1
Regards,

markbates:
There’s no flag by that name. If you run `buffalo g resource —help` it’ll show the available flags. I think what you’re looking for is `—skip-model`.

markbates:
This appears to be a regression from `v0.9.5`

jmbrunskill:
@markbates I checked out and build v0.9.5 tag from git.

It looks like the only difference between `buffalo g r users` and `buffalo g r users --use-model users` is that if you do not specify `--use-model` the generator will overwrite your 'users' model.

So if I am to understand the task correctly we need to.
A) Reinstate the --use-model flag
B) When the --use-model flag is passed in, don't over write the specified model file with a generated one

If so, I'm happy to have a go at implementing this. 
FYI - I'm pretty new to go (and to github contributions) but keen to learn.

markbates:
@jmbrunskill That's basically the gist of it. It basically turns off the call to the soda generator that generates a model and a migration. shouldn't be too hard to implement. Would love to see this happen! :)

jmbrunskill:
@markbates Great I've updated the code to handle this. 

A quick question before I submit a PR.
It looks like the old --use-model took a string which it used as the model name. 
I've replicated this functionality, but was wondering if it would be better to make --use-model a boolean and use the existing --name flag to set the model name?

jmbrunskill:
Decided to go ahead an create the PR can continue the discussion there.

PR #778, opened 2017-11-29
Fix Issue #751 - unknown flag: --use-model

jmbrunskill:
Fix for #751 - unknown flag: --use-model
Re-implemented the --use-model logic so that if this string is set, the generator will still generate code using the specified model but won't try to re-create the model or migration.

Also made some minor updates to improve readability in help text in `buffalo g r -h`


markbates:
@jmbrunskill Thanks so much. Just tried this out, works just as expected. Thanks for jumping on this regression. I was going to hold up the next release because of this issue. Huge help!

PR #789, opened 2017-12-12
Add UNIX domain socket binding support

stanislas-m:
* Allow Buffalo to run on an UNIX socket (e.g. unix:/tmp/buffalo.sock)

To test the feature (not available on Windows):
```
$ ADDR=unix:/tmp/buffalo.sock buffalo dev
```

```
$ curl --unix-socket /tmp/buffalo.sock http://localhost/
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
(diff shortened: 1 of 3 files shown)


markbates:
Awesome!

Issue #606, opened 2017-08-24
.env files support


dexafree:
As now [envy supports loading .env files](https://github.com/gobuffalo/envy/pull/3), I'd like to add Buffalo the capability to automatically load them.

A great way of integrating it to buffalo would be allowing env overriding by GO_ENV:

* Common ENV settings that won't vary across environments: `.env`
* Environment dependant settings (ie: SMTP_USER): `.env.development`, `.env.production`, `.env.test`

That is: `.env` file will always be loaded, but also `.env.<ENVIRONMENT>` if it exists. The latter will be able to override any values that are needed (if a user wants it).

I was thinking about putting it on `buffalo.Automatic`, as `buffalo.New` allows for complete custom tailor made apps, and not every user that decides to go that path may want .env file support.

I also thought about putting it in `optionsWithDefaults`, but as it's called also in `buffalo.New` I don't know if it would be appropriate.

As I mentioned on the envy repo, take into account that globally initialized vars won't be affected by this, as the resolution of the value would be made at the start of the program, and that would imply that the env files have not been loaded yet. This could affect to variables like this:

```go
const PORT = envy.Get("PORT", "8080")
```

As this kind of values are resolved before any other call has been made (including the envy.Reload one).
But this cannot be solved by any means with any library, as it's a language behaviour.

Before starting to code it, I'd like for the proposal to be commented and approved.

u007:
rails 5.1 recently added encrypted secret yml file. will be nice if we have this, then we dont need .env ?

```
bin/rails secrets:setup# generates config/secrets.yml.key and config/secrets.yml.enc
bin/rails secrets:edit# to edit config/secrets.yml and auto generate config/secrets.yml.enc
```

for more info:
http://www.starkandwayne.com/blog/rails-5-1-applications-can-be-a-lot-more-secretive-on-cloud-foundry-and-heroku/

paganotoni:
@markbates what do you think about implementing application secrets ? should we go initially with .env support or go for secrets.yml?

markbates:
I would need to see a proposal on how secrets would work and how they're different from a `.env` file.

paganotoni:
Yeah, I don't have it very clear but would love to provide `.env`

paganotoni:
@markbates i would love to get your feedback on #792 which is related to this one.

paganotoni:
@markbates An addition to #792 would be to generate a SESSION_SECRET in the `.env` generated file WDYT?

[...]

PR #792, opened 2017-12-14
Adds dot env support

paganotoni:
This PR adds `.env` support for generated buffalo apps, when generating a new buffalo app this adds an empty `.env` file, adds that file to the `.gitignore` and uses `envy.Load()` to load the `.env` every time the application restarts.

This PR also adds .env to the list of extensions that will be listened by refresh to restart the app, this is because it is cool to have the env change applied without stopping buffalo dev.

This is related with #606


generators/newapp/templates/actions/app.go.tmpl
```diff
@@ -34,6 +34,9 @@ var T *i18n.Translator
 // should be defined. This is the nerve center of your
 // application.
 func App() *buffalo.App {
+  //Loading .env file values
+  envy.Load()
```
markbates: Move this into `buffalo.New` that way it'll just work for every app, and they don't need the extra line. Check to make sure that moving this doesn't cause any problems for older apps, as well as apps that don't have a `.env`.

paganotoni:
@markbates thanks for the feedback, I changed the `eny.Load()` to be done in the buffalo.New and added a short description in the generated .env

paganotoni:
@markbates what do you think about generating the SECRET_KEY on the .env by default ?

markbates:
I don't think it's good idea. In development that key doesn't really matter, it only matters in production. From my understanding of the .env file is that it's not supposed to be committed, but is for each developer to set up for their environment. If that is correct we should also add it to the default .gitignore

paganotoni:
Yes, it is ignored currently, right, was thinking developers to have environment as closer to production as possible, but you're right if it doesn't make a difference then I'll forget it 👍

paganotoni:
Also, let me know if there is something else to do on this one and I'll jump into it, thanks for all the feedback!

PR #793, opened 2017-12-14
Minor Refactorings

paganotoni:


mail/smtp_sender.go
```diff
@@ -10,31 +10,51 @@ import (
 
 //SMTPSender allows to send Emails by connecting to a SMTP server.
 type SMTPSender struct {
-	Dialer *gomail.Dialer
+	Dialer  *gomail.Dialer
+	message *gomail.Message
```
stanislas-m: No sure this is a good idea to have the message here. How do you handle two concurrent `Send`?

paganotoni:
@stanislas-m thanks the heads up, you're right we better keep it as it was :)

buffalo/cmd/info.go
```diff
@@ -30,102 +30,74 @@ var infoCmd = &cobra.Command{
...
-			bb.WriteString("SQLite Not Found\n")
-		}
+}
 
-		return nil
-	},
+func execIfExists(name string, pathName string, c *exec.Cmd) {
+	bb := os.Stdout
+
+	if _, err := exec.LookPath("mysql"); err != nil {
```
markbates: `mysql` is hard coded here, which I'm assuming is not right.

buffalo/cmd/info.go
```diff
@@ -30,102 +30,74 @@ var infoCmd = &cobra.Command{
...
-			c.Stderr = bb
-			c.Run()
-		} else {
-			bb.WriteString("SQLite Not Found\n")
-		}
+}
 
-		return nil
-	},
+func execIfExists(name string, pathName string, c *exec.Cmd) {
```
markbates: You can clean up this function by passing in the `infoCommand` struct instead of all these args.

buffalo/cmd/info.go
```diff
@@ -30,102 +30,74 @@ var infoCmd = &cobra.Command{
...
-		bb.WriteString("\n### Dep Version\n")
-		if _, err := exec.LookPath("dep"); err == nil {
-			c = exec.Command("dep", "version")
-			c.Stdout = bb
-			c.Stderr = bb
-			c.Run()
-		} else {
-			bb.WriteString("dep Not Found\n")
-		}
+	bb.WriteString("\n### Go Version\n")
```
markbates: Why not make this part of the bigger loop below?

buffalo/cmd/info.go
```diff
@@ -30,102 +30,74 @@ var infoCmd = &cobra.Command{
...
-		bb.WriteString("\n### Dep Status\n")
-		if _, err := exec.LookPath("dep"); err == nil {
-			c = exec.Command("dep", "status")
-			c.Stdout = bb
-			c.Stderr = bb
-			c.Run()
-		} else {
-			bb.WriteString("dep Not Found\n")
-		}
+	bb.WriteString("\n### Go Env\n")
```
markbates: Why not make this part of the bigger loop below?

paganotoni:
@markbates All set on this one

markbates:
The only issue I've found with this is when I ran it against a project that didn't have `Gopkg.toml` file it returned an error status code, which isn't correct. It shouldn't return an error if one of things goes bad.

```
### Dep Status
no Gopkg.lock found. Run `dep ensure` to generate lock file
Usage:
  buffalo info [flags]

Flags:
  -h, --help   help for info

Error: exit status 1
```

paganotoni:
🤔 I just tried it with a newly created buffalo app (without `Gopkg.toml`) and did get :

```
### Dep Status
could not find project Gopkg.toml, use dep init to initiate a manifest
```

PR #808, opened 2017-12-18
added easy support for file uploads

markbates:


markbates:
https://vimeo.com/247866894

Issue #795, opened 2017-12-14
Buffalo mail package is using an unmaintained dependency


stanislas-m:
Using `dep ensure` with a Gopkg.toml file:
```
[[constraint]]
  branch = "development"
  name = "github.com/gobuffalo/buffalo"
```
The `gomail.v2` version fetched is the one from the **2.0.0** tag. This version API is not the one used by the Buffalo mail package (v2 branch), which leads to the following errors:

```
vendor/github.com/gobuffalo/buffalo/mail/smtp_sender.go:28: undefined: gomail.SetPartEncoding
vendor/github.com/gobuffalo/buffalo/mail/smtp_sender.go:28: too many arguments in call to m.SetBody
    have (string, string, <T>)
    want (string, string)
vendor/github.com/gobuffalo/buffalo/mail/smtp_sender.go:34: undefined: gomail.SetPartEncoding
vendor/github.com/gobuffalo/buffalo/mail/smtp_sender.go:34: too many arguments in call to m.AddAlternative
    have (string, string, <T>)
    want (string, string)
vendor/github.com/gobuffalo/buffalo/mail/smtp_sender.go:77: dialer.Username undefined (type *gomail.Dialer has no field or method Username)
vendor/github.com/gobuffalo/buffalo/mail/smtp_sender.go:78: dialer.Password undefined (type *gomail.Dialer has no field or method Password)
===
```

I think we can fix the `dep` problem, but it seems `gomail.v2` package is not maintained anymore. We should use a fork: `mail.v2` is a first guess, but it seems they have/had recent management issues.

paganotoni:
Covered by #812

PR #812, opened 2017-12-19
Moves to use gopkg.in/mail.v2

paganotoni:
This is related to #795, so far it seems not to break dependencies, but we will have to test it more extensively.

### DIFF
--- a/mail/smtp_sender.go
+++ b/mail/smtp_sender.go
@@ -5,7 +5,7 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
-	gomail "gopkg.in/gomail.v2"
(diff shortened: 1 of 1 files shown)


paganotoni:
Just tested this by doing the following:

- buffalo new app
- buffalo g mailer welcome
- changing properties
- calling `mailers.SendWelcome()` from my home handler

Do you think that is ok to say moving it to mail.v2 is ok?

markbates:
lol, yeah, it definitely feels too easy. :)

@stanislas-m would you mind giving this a go since you found the problem originally?

stanislas-m:
@markbates Sure, I'll test this one today. :)

Issue #797, opened 2017-12-15
Default Robots.txt Handler


paganotoni:
One small addition to the buffalo framework that I would like to do is have the buffalo apps handle [`robots.txt`](https://en.wikipedia.org/wiki/Robots_exclusion_standard) by default.

**How will this work**

We will add a default robots handler to buffalo apps, what this handler will do is will render a file in `robots.txt`  that is inside `/assets/` folder.

If the file is not present it will render a permissive `robots.txt` content

```
User-agent: *
Disallow:
```

We should also allow developers to change this handler for any other, something like:
```
app.RobotsHandler = otherRobotHandler
```

**What needs to change**

- Need to generate a default `robots.txt` inside assets folder.
- Need to add a default `robots.txt` handler that will render the file from the assetsBox.
- Need to add documentation of all this on `gobuffalo.io`

u007:
would be better a way to render any files in public/ if it exists without the need to add prefix assets

and also anyway to off directory indexing for app.ServeFiles ?

markbates:
@paganotoni I like @u007 idea of allowing for the `public` directory to be served up easily. #853 addresses this.

u007:
thank you :)

PR #819, opened 2017-12-21
Adding robots handler

paganotoni:
This is related to #797


generators/assets/webpack/templates/assets/robots.txt.tmpl
```diff
@@ -0,0 +1,2 @@
+User-agent: *
```
markbates: We should probably generate this file for non-webpack apps as well.

paganotoni: Right, just added that default robots to the non-webpack generated app.

paganotoni:
@markbates thoughts on this one?

markbates:
I'm still thinking about. Not sure if this is how we should address the problem. It hasn't escaped my attention, but I need some more time to mull it over.

markbates:
@paganotoni I'm closing this in favor of #853 which allows for a more generic serving up of the `public` folder.

paganotoni:
Sounds good, I can see that my solution was too specific to the robots issue, thanks @markbates

PR #820, opened 2017-12-21
Adds favicon to newly generated buffalo apps

paganotoni:


generators/newapp/templates/templates/application.html.tmpl
```diff
@@ -9,6 +9,7 @@
     <%= stylesheetTag("application.css") %>
     <meta name="csrf-param" content="authenticity_token" />
     <meta name="csrf-token" content="<%= authenticity_token %>" />
+    <link rel="icon" href="/assets/images/favicon.ico">
```
markbates: This should probably use the `assetPath` helper:

```
<link rel="icon" href="<%= assetPath("images/favicon.ico") %>">
```

paganotoni:
good point, I changed it to use `assetPath`

PR #834, opened 2018-01-08
feat: use a checkbox when the type is bool

josegonzalez:


generators/resource/templates/templates/model-view-_form.html.tmpl
```diff
@@ -2,8 +2,12 @@
 {{ if eq $p.Type "text" -}}
 <%= f.TextArea("{{$p.Name}}", {rows: 10}) %>
 {{ else -}}
+{{ if eq $p.Type "bool" -}}
+<%= f.CheckboxTag("{{$p.Name}}") %>
```
markbates: This line should probably be `<%= f.CheckboxTag("{{$p.Name}}", {unchecked: false}) %>`. This way it will send a `false` value when unchecked, instead of nothing.

markbates:
Thanks!

Issue #845, opened 2018-01-10
NODE_ENV not being set to production during build


bastiankoetsier:
With the [removal](https://github.com/gobuffalo/buffalo/commit/436dc23f762ecc9aef1c759db2e7c09193944414#diff-d8610c51e98c7aa8d24252f5dbec0bfeR25) of loading the whole environment-variables before `exec`, it is prevented  that `NODE_ENV` gets actually [set](https://github.com/gobuffalo/buffalo/blob/master/buffalo/cmd/build/assets.go#L13) to `production`  during `buffalo build` .

This prevents the execution of webpack plugins like uglify for a production build.
It can be circumvented by setting `NODE_ENV=production buffalo build`

Zarkopafilis:
Probably fixed by https://github.com/gobuffalo/buffalo/pull/846 (if it gets merged)

Issue #841, opened 2018-01-09
Allow for the binary's default environment to be set during build time


markbates:
When running `buffalo build` there should be a `--environment` flag that let's the user set the default environment of the binary from `development` to whatever they would like.

PR #846, opened 2018-01-11
"Fix" Issue #841

Zarkopafilis:
I want to make sure if this should be the proper way to set the development environment for the build, in order to procceed with testing & possible mistake corrections.

A new environment flag is registered and defaulted to "development".
Then, on **bin.go**, its passed to envy (not persisted to the underlying environment), in order to procceed to **exec.go** without changing function signatures.

The value is retrieved from envy and appended in the Env slice of **exec.Cmd** so it can be used instead of the default one on the actual command execution step.

https://github.com/gobuffalo/buffalo/issues/841

### DIFF
--- a/buffalo/cmd/build.go
+++ b/buffalo/cmd/build.go
@@ -77,5 +77,5 @@ func init() {
 	xbuildCmd.Flags().StringVar(&buildOptions.LDFlags, "ldflags", "", "set any ldflags to be passed to the go build")
 	xbuildCmd.Flags().BoolVarP(&buildOptions.Debug, "debug", "d", false, "print debugging information")
 	xbuildCmd.Flags().BoolVarP(&buildOptions.Compress, "compress", "c", true, "compress static files in the binary")
-	xbuildCmd.Flags().StringVarP(&buildOptions.Environment, "environment", "e", "development", "set the environment for the binary (defaults to 'development')")
+	xbuildCmd.Flags().StringVarP(&buildOptions.Environment, "environment", "", "development", "set the environment for the binary")
 }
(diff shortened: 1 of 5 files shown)


markbates:
Thanks for looking into, unfortunately it doesn't solve the issue of #841. The behind that ticket is the binary that built from running `buffalo build` can be set to run in an environment other than development. This means the binary needs to get the environment "burnt" into it. This PR simply sets the requested environment of the `buffalo build` command, and unfortunately, not the finish binary. Does that make sense?

Also, this line:

```go
cmd.Env = envy.Environ()
```

breaks Windows, which is why it was removed and caused the problem of #845. When implemented #841 won't fix #845 unfortunately.

markbates:
The good news is, you're very close to the actual implementation. :) I was thinking that adding a line to the `main.go` file that gets generated as part of the build process could contain something like:

```go
func init() {
  envy.Set("GO_ENV", "<environment>")
}
```

That's might be enough to do it.

Zarkopafilis:
Should it be just `Set` or `MustSet` (Try to persist to the host as well)?

markbates:
@Zarkopafilis I'm closing this PR in favor of #850 because GitHub won't let me push a commit to this PR.

When testing this I found that it still wasn't working. I dug deeper and found that the place that we wanted to set the ENV var was too late. The `actions` package was already loaded by that point, so the `actions.ENV` was already set.

In order to get it to load properly, I moved the setting of the ENV var from the `main` package to the `a` package which gets loaded __before__ `actions`. That solved the problem. :)

Thanks for taking point on this.

Issue #847, opened 2018-01-12
Webpack should compile different JS files


paganotoni:
Under some situations as a developer you want to have different JS for different pages in your application, this can be for multiple reasons, the reasons I've faced are:

- You don't want to load unneeded libraries on every page
- You don't want to expose some JS functionality in public pages
- You want to have your JS logic separated

As a solution Buffalo could have webpack compile every file on the /assets/js folder into its fingerprinted version inside `/public/assets` and in order to avoid partial files, we could exclude files that start with an underscore or another pattern we decide.

Changes to make this possible would be done in the generated `webpack.config.js`.

markbates:
Love it. I would use `_file.js` for partials to match plush and sass.

paganotoni:
Great! it seems that for our sass files it's the same story, we're only generating `application.css`, should we include in this issue doing the same with sass (i can see almost the same reasons to separate your style).

markbates:
I assumed you meant both. :) Definitely!

stanislas-m:
@paganotoni How about this? https://github.com/webpack/webpack/tree/master/examples/multiple-entry-points

paganotoni:
@stanislas-m #849

markbates:
@paganotoni I started working on a new application and found out that this is causing some problems. 

Everything works fine, but when I change, say `application.js` it cleans out the old `application.hash.js` file, but never rebuilds it and updates the `manifest.json`. This results in a lot of these types of errors:

```
could not find assets/application.7cf2e05554e16f6c5e3c.js
```

The only way to fix the problem is to restart `buffalo dev`. 

This either needs to be fixed before we release `0.11.0` or it needs to be reverted.

paganotoni:
@markbates when are we planning to release 0.11.0 ? I want to commit to have this in 0.11.0

markbates:
right now February 12th

paganotoni:
In that case let’s have it in `0.11.0`

paganotoni:
@markbates this issue seems not to be related with this particular ticket, anyways I'm working on it, I changed my webpack configuration to be the old one and it still happens when you change a file and save it multiple times quickly, it kind of loses contact with the assets folder and webpack stops recompiling.

There is another issue with the fonts that I've added in #821.

I continued looking into the config and seems the culprit is the `CleanWebpackPlugin` configuration we're using, by disabling the `watch` property this seems to be fixed, I'm preparing a PR for this and #821 soon.

PR #849, opened 2018-01-12
Multiple output scss js files

paganotoni:
Allows having multiple output `SCSS` and `JS` files in our buffalo apps by default.
Covers #847.


paganotoni:
@markbates I think is working, let me know how is it looking for you once you have the chance to test it.

PR #858, opened 2018-01-13
Validate template syntax when building binary fixes #840

### LINKED ISSUE #840: Validate template syntax when building binary
When running `buffalo build` we should validate that the plush templates are syntactically valid.

markbates:


yatish27:
❤️ 👍

PR #913, opened 2018-02-09
Revert "Allow for testing a single package"

markbates:
Reverts gobuffalo/buffalo#912


arschles:
@markbates is there another way to test an individual package? Not sure exactly how the testify `-m` support works, so feel free to punt me to those docs if that's the place to look.

PR #932, opened 2018-02-18
webpack/style_regex: change regex to catch both .sass and .scss files

joncarr:


joncarr:
Hi @paganotoni  thanks for checking it out!  I'm not sure what you mean about having multiple entries? Can you please elaborate and I'll get the changes taken care of.

joncarr:
I just threw together this small sample project to demonstrate (without pop) -- sorry, maybe I should have done this previously..|
https://github.com/joncarr/BuffaloStylecheck

paganotoni:
Sure, yes, so, previously we added the possibility to have multiple scss compiled files, this is in case you wanted to have page-specific stylesheets so on line 16 of the same webpack template file we added any file not starting with an underscore to be compiled to CSS by webpack (added as an entry).

I think we need to allow `.sass` files to be added as well and in that same line, you will notice that function only takes into consideration `.scss`.

joncarr:
Oh, My apologies, I see that now!  My mistake...one sec :)

paganotoni:
nothing to worry, thanks for this PR! it was something I've missed before, probably because i just use .scss.

joncarr:
I also work with mainly scss....I have just run across situations where I needed something quick, or was provided styling that contained sass.   Rather than going through, I just changed my webpack, now it's just a habit... :)

Issue #928, opened 2018-02-17
Docker-compose generator


stanislas-m:
Since Buffalo already generates a Docker file, it can be a great addition to generate a docker-compose.yml file too.

This file will be generated from the DB params provided to the `new` command.

Here is a raw template for a postgres version (just an idea, not the final thing):

```
version: '3'

services:
  db:
    container_name: {{ .opts.Name.Title }}-db
    image: postgres
    environment:
      POSTGRES_PASSWORD: postgres
      POSTGRES_DB: {{ .opts.Name.Title }}_${GO_ENV:-development}
    volumes:
      - ./db:/var/lib/postgresql/data
```

markbates:
more color:

What I would love would be a generator, or a flag on new, that sets up docker compose with the app and one of the support dbs all configured to work together out of the box. That would be really nice.

Then buffalo dev could check for all that and start it up and run the app locally in the docker container.

gintsgints:
We would need script like this:

```
until nc -z ${MYSQL_HOST} ${MYSQL_PORT}; do
    echo "$(date) - waiting for mysql..."
    sleep 1
done
```
In docker image. For that to make, how about to move helper docker image sources out to different project?

gintsgints:
I made docker image with script included gintsgints/docker-buffalo , but sadly when I do 

```
buffalo dev
```
with generated app inside that container, it stacks after font-awesome.

gintsgints:
Got it working for dev environment. So, bit updated Dockerfile (based on my modified image & put ADDR env variable also for builder build. Begining:

```
# This is a multi-stage Dockerfile and requires >= Docker 17.05
# https://docs.docker.com/engine/userguide/eng-image/multistage-build/
FROM gintsgints/docker-buffalo as builder

RUN mkdir -p $GOPATH/src/gitlab.com/gintsgints/coke
WORKDIR $GOPATH/src/gitlab.com/gintsgints/coke

# this will cache the npm install step, unless package.json changes
ADD package.json .
ADD yarn.lock .
RUN yarn install --no-progress
ADD . .
RUN go get $(go list ./... | grep -v /vendor/)
RUN buffalo build --static -o /bin/app

# Bind the app to 0.0.0.0 so it can be seen from outside the container
ENV ADDR=0.0.0.0
```

And compose file as follows:

```
version: '3.5'
services:
  app:
    build:
      context: .
      target: builder
    depends_on:
    - db
    ports:
    - "3000:3000"
    volumes:
[...]

PR #951, opened 2018-03-02
Creation of compose file (#928)

gintsgints:
First step towards compose support. This should work for postgres & mysql.

To start development environment do
```
docker-compose -f docker-compose.dev.yml build
docker-compose -f docker-compose.dev.yml up -d
```


markbates:
I haven’t played with this yet, but my first take is it needs cockroachdb support. SQLite doesn’t need it. Also, does it make sense to only generate this file if it’s anything other than SQLite? Finally, it would be great if buffalo dev picked up this file and started it all for you. Just some initial thoughts.

gintsgints:
I will work on cockroachDB support now.

gintsgints:
> Running the docker-compose file, the buffalo app doesn't start

Yeah... I forgot, I did change app base image. To try how it works you have to update app image to:

```
FROM gintsgints/docker-buffalo as builder
```

That script - wait-for, is only addition to normal app image.

gintsgints:
I would recommend to stay with application built within compose file. Because:

- one still will have his development files outside docker for editing (they are mounted as volume)
- you do not need rebuild image usualy while in development
- in case you still need to rebuild app docker, it takes not much time
- same compose file can be used for deployment

gintsgints:
Added cockroachDB suport.

robbyoconnor:
Can we name this `docker-compose.yml`?

generators/docker/multi/templates/Dockerfile.tmpl
```diff
@@ -1,6 +1,6 @@
 # This is a multi-stage Dockerfile and requires >= Docker 17.05
 # https://docs.docker.com/engine/userguide/eng-image/multistage-build/
-FROM gobuffalo/buffalo:{{.opts.Version}} as builder
+FROM gintsgints/docker-buffalo as builder
```
robbyoconnor: ??

gintsgints: I propose we have different project for making docker app envionrment, as it will contain files which needed just for docker environment. As example script - wait-for.  It will have seperate versioning.
If I could create project under gobuffalo I would make that image there.

robbyoconnor: Why `docker-buffalo` as the name...also I don't wanna rely on your image being up to date...so yeh -- let's NOT do this. I do semi-agree -- but at the same time -- it might be worth it to host the image as a buffalo repo.

gintsgints: As soon as there will be some docker image which contains wait-for script, we can replace reference back. Other way my supplied docker-compose will not work at all.

gintsgints: Please give me place where I could put my lovely script :)

gintsgints: One more reason, to have seperate project for buffalo docker image is, we need some variations for base image, like one for arm64.

robbyoconnor:
Shouldn't the docker-compose file be optimized for development and run `buffalo dev`?

robbyoconnor:
Mount the project code as a volume

gintsgints:
- project code already should be mounted

```
    volumes:
    - .:/go/src/{{.opts.App.PackagePkg}}
```
- buffalo dev will follow after we clarify fix Docker base image.

gintsgints:
Docker compose version 3.5 is needed, to have option choose first build from multistage builds.

markbates:
@gintsgints @stanislas-m what's the status of this PR?

stanislas-m:
@markbates I'll take a look soon.

PR #953, opened 2018-03-03
Allow "go test" to work without requiring GO_ENV to be set to "test"

RedPh0enix:
Fix proposed by TankTheFrank


RedPh0enix:
Apologies, should have gone to development, rather than master.

markbates:
@RedPh0enix I think this is a better solution to fixing it https://github.com/gobuffalo/envy/pull/10

RedPh0enix:
Agree - much better to have it solved in the imported packages, rather than in the generated code.

Issue #930, opened 2018-02-17
Default Production Error page is ugly.


markbates:
https://www.evernote.com/l/ABGjR5egIT9CL6QOL6bImIMVFRk96Nuluek

It would be nice to have, even just a little bit, of styling on that page.

paganotoni:
@markbates i can come up with something simple here quick, I'll keep you posted.

stanislas-m:
I'd love to have a "buffalo skull" version of the logo, for these pages.

Something like that:
![image](https://user-images.githubusercontent.com/1911803/36344213-382f0fb0-1417-11e8-8df3-fa017ef07718.png)

AlhasanIQ:
i would suggest using something like the error pages included in the last version of laravel

PR #956, opened 2018-03-06
Less ugly production error page

paganotoni:
Given what we discussed in #930 I decided to start with this simple design for the error page, it adds some styling to it so we can have a better-looking error page for production:

error-page

(nothing fancy for now)

@markbates @stanislas-m sorry for the delay with this one, I think we can iterate from this first version, LMK if you guys want to change copy or any style you don't like.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -220,11 +220,75 @@ var devErrorTmpl = `
 </html>
 `
 var prodErrorTmpl = `
-<h1>We're Sorry!</h1>
-<p>
-It looks like something went wrong! Don't worry, we are aware of the problem and are looking into it.
-</p>
-<p>
-Sorry if this has caused you any problems. Please check back again later.
-</p>
+<!DOCTYPE html>
+<html>
+<head>
+<style>
+body {
+	background: #ECECEC;
+	padding-top: 25px;
+}
+
+.card {
+	box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


markbates:
Considering that looks a million times better than what we have now, I’m happy!

stanislas-m:
@paganotoni That's way better! :+1:  Is it possible to have a different text, at least for the 404 Not Found page? "Something went wrong" is not the kind of error you're used to see for a page not found, in my head it sounds I've broken something.

paganotoni:
@stanislas-m i like that idea, i'll add that into this PR, agree that its a bit odd to see an error when its actually a 404.

paganotoni:
@stanislas-m  not-found

stanislas-m:
@paganotoni I'm not sure the sentence about the app owner is necessary, but overall that's great! (404 errors are really common, and as an app owner I'll probably look at the logs for server errors, not 404 ones.)

paganotoni:
@stanislas-m @markbates I'm removing the WIP, added another flavor for 404 as @stanislas-m mentioned, give it a shot and let me know :)

robbyoconnor:
I vote we use https://http.cat photos :D

paganotoni:
@markbates this one is ready to be merged.

PR #957, opened 2018-03-06
Refactor i18n middleware language finders

stanislas-m:
* Split language finders into separate functions
* Refactor LanguageFinder type
* Generalize Language finders options
* Add a new path prefix language finding strategy
* Ensure setups with default config won't break

Related to: https://github.com/gobuffalo/gobuffalo/issues/133


markbates:
@stanislas-m this appears to be a breaking change, how do you propose handling that? via `buffalo update`?

stanislas-m:
@markbates `buffalo update` seems to be the obvious way, yeah. This should only affect projects with a customized i18n middleware (I don't think they are legion), but we can at least give some help for those cases.

markbates:
The bare minimum we add to `buffalo update` should at least detect they’re using stuff that changed and print a warning that they should change things. The nice to have would be doing the code changes for them, if possible.

stanislas-m:
`buffalo update` should take care of these cases:

```go
T.CookieName = "language" // Becomes T.LanguageFinderOptions["CookieName"] = "language"
T.SessionName = "language" // Becomes T.LanguageFinderOptions["SessionName"] = "language"
// This one needs refactor:
T.LanguageFinder = func(t *i18n.Translator, c buffalo.Context) []string {
	langs := make([]string, 0)
	if cookie, err := c.Request().Cookie(t.CookieName); err == nil {
		if cookie.Value != "" {
			langs = append(langs, cookie.Value)
		}
	}
	langs = append(langs, t.DefaultLanguage)
	return langs
}
```

markbates:
@stanislas-m can you make the changes needed to `buffalo update` to make sure this doesn't affect too many people?

stanislas-m:
@markbates Sure, it's on my todo list. :)

stanislas-m:
@markbates The update works on my side, but a specific case like the following will break:

```go
T.LanguageFinder = func(t *i18n.Translator, c buffalo.Context) []string {
	langs := make([]string, 0)
	if cookie, err := c.Request().Cookie(t.CookieName); err == nil {
		if cookie.Value != "" {
			langs = append(langs, cookie.Value)
		}
	}
	langs = append(langs, t.DefaultLanguage)
	return langs
}
```

I issue a warning for this one with the update command, is it enough?

middleware/i18n/i18n.go
```diff
@@ -14,27 +14,32 @@ import (
 	"github.com/pkg/errors"
 )
 
-// LanguageFinder can be implemented for custom finding of search
+// LanguageFinder - deprecated
```
markbates: what should they use instead?

stanislas-m: I just added more infos in the last commit, thanks!

PR #958, opened 2018-03-06
Added viewport meta tag to header to make it responsive

deanroker123:


paganotoni:
Thanks! merging this one.

PR #961, opened 2018-03-10
Fix broken URL in SHOULDERS.md

ruudk:


markbates:
I'm merging this, because I want you to get contributor credit, but that file is actually auto generated by https://github.com/gobuffalo/buffalo/blob/master/grifts/shoulders.go, which is run when a new release is cut, so any changes to the SHOULDERS.md that are manually made will get wiped out. I would love a PR that fixed this automatically, if you're interested. :)

ruudk:
I see the issue now, it's impossible to fix. 

Only solution I see is this: When url is from github.com then split url and insert `/tree/master/` between path 2 and 3 for the URL. Not sure what you think about it?

But then I think it would be better to just change the URL to always point to `https://godoc.org/xxx` instead.

ruudk:
Cool, created PR in https://github.com/gobuffalo/buffalo/pull/964

PR #962, opened 2018-03-10
Add newline after `db create`

ruudk:
When creating a new project, I found that the generated `README.md` was missing a newline after the `buffalo db create -a` command.

See:
```md
...
### Create Your Databases

Ok, so you've edited the "database.yml" file and started postgres, now Buffalo can create the databases in that file for you:

	$ buffalo db create -a
## Starting the Application
...
```

This PR adds an extra newline after the command to make it more pretty!


markbates:
thanks!

PR #968, opened 2018-03-16
Adds PerformEvery to simple worker.

AaronVasquez:


AaronVasquez:
Didn't want to bring in a new dependency just to copy a struct, but not sure of the best way to do so in Golang.

PR #969, opened 2018-03-16
Adds performEvery to worker

AaronVasquez:


markbates:
I don't think this is the correct approach to the problem. 

First, it changes the https://godoc.org/github.com/gobuffalo/buffalo/worker#Worker interface, that means that every implementation of that interface would be broken. When breaking APIs/interfaces, it's important to give developers a long enough run way that they can make the necessary changes.

Also, I'm not sure this is something that is necessary. You could just re-enqueue the job at the end of each run:

```go
func someWorkerFunc(args worker.Args) error {
	w := App().Worker
	job := worker.Job{
		Queue:   "default",
		Handler: "someWorkerFunc",
		Args:    args,
	}
	w.PerformIn(job, job.Args["duration"].(time.Duration))
	return nil
}

func init() {
	w := App().Worker
	job := worker.Job{
		Queue:   "default",
		Handler: "someWorkerFunc",
		Args: worker.Args{
			"duration": 15 * time.Minute,
		},
	}
	w.PerformIn(job, job.Args["duration"].(time.Duration))
}
```

PR #975, opened 2018-03-16
Commit Gopkg.lock and remove vendored blackfriday 

### LINKED ISSUE #973: Commit Gopkg.lock and remove vendored blackfriday
Hi,

Following up the discussion on Slack, here's a couple of things that I've noticed that need to be addressed before I can contribute to buffalo:

- https://github.com/gobuffalo/buffalo/tree/master/vendor/github.com/russross/blackfriday -> this is the only committed dependency. Is there any reason for this to be committed?
- ` Gopkg.lock ` is not committed so my version of the dependencies might be different than someone else's version.

Thank you.

markbates:
fixes #973


dlsniper:
There is something to consider about the ` Gopkg.lock ` I've forgot to mention in the issue. What happens when using the default get method without sqlite, and what happens when using sqlite?

markbates:
Since `go get` doesn’t care about the lock stuff, is that question moot?

PR #977, opened 2018-03-17
Check for dep and goimports before go getting.

barrongineer:
Conditionally `go get`ing `dep` and `goimports`. I noticed that even though I had them installed via package manger on Arch, Buffalo was `go get`ing them. Might be annoying for those that like a clean `GOPATH`.


markbates:
Thanks!

Issue #982, opened 2018-03-22
Invalid case-insensitive filepath handling


egonelbre:
Some FileSystems are case-insensitive, mainly Windows, which means using `strings.TrimPrefix` to find the relative directory or path, will eventually fail. Something like this, will be probably more successful in most cases:

``` go
func ImportPath(gopaths []string, targetpath string) (string, error) {
	var err error
	var rel string
	for _, gopath := range gopaths {
		rel, err = filepath.Rel(gopath, targetpath)
		if err == nil {
			return filepath.ToSlash(rel), nil
		}
	}
	return "", err
}
```

Example in Windows with a case-mismatch:

```
f:\Go\src\sandbox\buffalo>echo %GOPATH%
F:\Go

f:\Go\src\sandbox\buffalo>buffalo new example
      create  .buffalo.dev.yml
      create  assets/images/logo.svg
      create  assets\css\application.scss
<<< SNIP >>>
      create  public\robots.txt
      create  templates\_flash.html
      create  templates\application.html
      create  templates\index.html
         run  go get -t ./...
can't load package: package sandbox/buffalo/example:
main.go:6:3: invalid import path: "f:/Go/src/sandbox/buffalo/example/actions"
can't load package: package sandbox/buffalo/example/actions:
actions\app.go:11:3: invalid import path: "f:/Go/src/sandbox/buffalo/example/models"
can't load package: package sandbox/buffalo/example/grifts:
grifts\init.go:5:2: invalid import path: "f:/Go/src/sandbox/buffalo/example/actions"
Usage:
  buffalo new [name] [flags]
<<< SNIP >>>
ERRO[0038] Error: exit status 1
```

And the main.go, app.go, Dockerfile... contain import paths such as:

``` go
package main

import (
  "log"

  "f:/Go/src/sandbox/buffalo/example/actions"
)

func main() {
  app := actions.App()
  if err := app.Serve(); err != nil {
    log.Fatal(err)
  }
}
```

I suspect there are other places that may contain similar mistakes.

markbates:
A PR would be very welcome!

-----------
Mark Bates

egonelbre:
Tried my best.

I wasn't sure what the policy of error handling is, but I took the following approach:
1. When there was error handling around, returned the error
2. When there was no error handling, then fell back to strings.TrimPrefix results.

[...]

PR #989, opened 2018-03-24
Fix relative path finding (#982)

egonelbre:
strings.TrimPrefix can give the wrong result with case-insensitive
paths. filepath.Rel usually works better.


markbates:
Thank you!

PR #994, opened 2018-03-26
Added http:// to server address in log to make it clickable

forrest321:
Enhancement: After starting a server, it would be nice to be able to click the server address in the console to open a browser. Adding http:// before the address in the current log output will make it clickable in some IDE's, and easier to copy & paste in others.

This is an edit to a line of logs, so it could be considered a content or documentation change.  I'm not sure there is an easily accessible way to write unit tests.


forrest321:
linkfix

markbates:
Brilliant. Thanks!

PR #998, opened 2018-03-29
Fix wording re. commenting/uncommenting parts of Dockerfile.

matt-harvey:
To "comment out" a piece of code is to make it _into_ a comment.

To "uncomment" something is to turn it from a comment into
executable code.


matt-harvey:
@markbates You're welcome; and thanks for the rapid turnaround.

Issue #997, opened 2018-03-28
Proposal: Live asset reloader 


jaynagpaul:
One of my favorite things from Phoenix framework is the [live asset reloader](https://github.com/phoenixframework/phoenix_live_reload). It would be great to have a live asset reloader during development. Here's the description of what it does taken straight from the readme.

> this project adds a plug which injects some JavaScript into your page with a WebSocket connection to the server. When you make a change to anything in your config for live_reload (JavaScript, stylesheets, templates, and views by default) then the page will be reloaded in response to a message sent via the WebSocket. If the change was to an Elixir file then it will be recompiled and served when the page is reloaded. If it is JavaScript or CSS, then only assets are reloaded, without triggering a full page load.

Wanted Features:
- Recompile templates
- Watch javascript / CSS and send a message via websocket to reload

I may attempt to implement this myself in the future.

JacobSoderblom:
My suggestion is to use [webpack-dev-server](https://webpack.js.org/configuration/dev-server/) on the `buffalo dev` command and  [write-file-webpack-plugin](https://github.com/gajus/write-file-webpack-plugin) as a plugin to `webpack config`.

Due to `webpack-dev-server` store the files in the memory and serve them through a node server, we will have to change how `buffalo` handle the assets. But if we use `write-file-webpack-plugin` we will be able to use the assets as they are today.

This will solve the _Watch javascript / CSS and send a message via websocket to reload_ feature.

There might be better ways to solve this, but I don't know if there are any easier ways.

markbates:
@JacobSoderblom I'd love to see a PR, would you be interested in taking a crack at it?

paganotoni:
Covered by #1027

PR #1004, opened 2018-04-04
Live asset reloading

JacobSoderblom:
Solves #997 

I was not able to use the latest `webpack-dev-server` that supports `writeToDisk` with `webpack` version `3.x`, so I had to use [`WriteFilePlugin`](https://github.com/gajus/write-file-webpack-plugin/pulls) instead, they both solves the same issue.

### DIFF
--- a/buffalo/cmd/dev.go
+++ b/buffalo/cmd/dev.go
@@ -92,7 +92,13 @@ func startWebpack(ctx context.Context) error {
 		}
 	}
 
-	cmd := exec.CommandContext(ctx, webpack.BinPath, "--watch")
+	if _, err := os.Stat(webpack.DevServerPath); err != nil {
+		if os.IsNotExist(err) {
+			return errors.New("webpack-dev-server is not installed, please install webpack-dev-server locally")
+		}
+	}
+
+	cmd := exec.CommandContext(ctx, webpack.DevServerPath)
 	cmd.Stdin = os.Stdin
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
--- a/generators/assets/webpack/templates/package.json.tmpl
+++ b/generators/assets/webpack/templates/package.json.tmpl
@@ -31,6 +31,8 @@
     "uglifyjs-webpack-plugin": "~0.4.6",
     "url-loader": "~0.6.2",
     "webpack": "~3.10.0",
-    "webpack-manifest-plugin": "~1.2.1"
+    "webpack-manifest-plugin": "~1.2.1",
+    "webpack-dev-server": "~2.11.1",
+    "write-file-webpack-plugin": "~4.2.0"
   }
 }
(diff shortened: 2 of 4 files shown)


buffalo/cmd/dev.go
```diff
@@ -92,7 +92,13 @@ func startWebpack(ctx context.Context) error {
 		}
 	}
 
-	cmd := exec.CommandContext(ctx, webpack.BinPath, "--watch")
+	if _, err := os.Stat(webpack.DevServerPath); err != nil {
```
markbates: Currently this isn't backward compatible. However, if you change it to the following, it should work with current `v0.11.0` applications.

```go
	bin := webpack.DevServerPath
	if _, err := os.Stat(bin); err != nil {
		logrus.Warnf("%s is not installed, please install webpack-dev-server locally, trying to fallback to %s", bin, webpack.BinPath)

		bin = webpack.BinPath
		if _, err := os.Stat(bin); err != nil {
			return errors.Errorf("could not find either %s or %s", webpack.DevServerPath, webpack.BinPath)
		}
	}

	cmd := exec.CommandContext(ctx, bin)
	cmd.Stdin = os.Stdin
	cmd.Stderr = os.Stderr
	cmd.Stdout = os.Stdout
	return cmd.Run()
```

buffalo/cmd/dev.go
```diff
@@ -92,7 +92,13 @@ func startWebpack(ctx context.Context) error {
...
+		if os.IsNotExist(err) {
```
markbates: for me the error that was returned was:

```go
### err -> &os.PathError{
    Op:   "stat",
    Path: "node_modules/.bin/webpack-dev-server",
    Err:  syscall.Errno(0x2),
}
```

I think it's safe enough to say any error returned from `os.Stat` should be good enough.

buffalo/cmd/dev.go
JacobSoderblom: Yea, you're right. I'll fix it!

markbates:
@JacobSoderblom there are some conflicts that need resolving, can you please resolve them so I can test. thanks?

JacobSoderblom:
@markbates yes, sorry for the delay.

markbates:
Any idea what `(node:14308) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead` means?

markbates:
@JacobSoderblom I don't know what happened, but a bad merge happened, and now I can't tell GitHub to re-open this PR. Sorry. Can you re-open another one with my earlier comments:

> This seems to work. There are a couple of issues though.
>
> The first is we've seem to have lost the web pack output, so I had no idea if my assets actually compiled or not, including web pack errors. That definitely needs to be fixed.
>
> The other thing I would like to see is if i change a template, I feel like that should also trigger a live reload.

JacobSoderblom:
@markbates No problem!

Sure, but it can take a little while. I don't have that much time right now for open source, but I'll try to fix it as soon as possible!

PR #1006, opened 2018-04-04
Add test for template partial rendering with a local variable.

ajcollins:
Hi Mark,

After reading the documentation on partial rendering (https://gobuffalo.io/en/docs/partials#local-context) I expected local and global variables to be merged. This doesn't seem to happen at present. That may be the desired behaviour?

This adds a failing test case. It could be combined with the existing test case, depending on your preferred approach.

I've created a candidate fix in `plush/compiler.go` for which I'll submit an MR for review.

Best wishes,

Alex

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,4 +1,4 @@
-FROM golang:1.9.0
+FROM golang:latest
 
 RUN go version
 
--- a/render/template.go
+++ b/render/template.go
@@ -20,6 +20,7 @@ type templateRenderer struct {
 	*Engine
 	contentType string
 	names       []string
+	data        Data
 }
 
 func (s templateRenderer) ContentType() string {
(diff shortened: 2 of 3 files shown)


markbates:
I think the correct fix is here in the buffalo repo, instead of the plush repo, since the `partial` helper is defined in buffalo. https://github.com/gobuffalo/buffalo/blob/master/render/template.go#L43

ajcollins:
That was what I'd originally hoped, but I couldn't see a way for the partial helper to access the required context information: since plush is responsible for binding the arguments to the function call. Perhaps I'm missing something?

Happy to update the PR if you can suggest where to access this (or change how it's accessed).

markbates:
I think you need to start here https://github.com/gobuffalo/buffalo/blob/master/render/template.go#L29 capture the original `Data` and hold onto it in the struct. Then in `partial` you would want to create a new map that merges the renderer Data and the partial's data.

ajcollins:
Snap! I'll do that then - thanks.

PR #1007, opened 2018-04-05
Moving to use Webpack 4

paganotoni:
This PR moves new buffalo templates to use webpack 4.5 it also attempts to organize the JS code in generated webpack.config.js.

Props on this one to @stanislas-m for all the help and guidance!


generators/assets/webpack/templates/webpack.config.js.tmpl
```diff
@@ -1,117 +1,104 @@
...
+const configurator = {
+  entries: function(){
+    var entries = {
+      application: [
+        './node_modules/jquery-ujs/src/rails.js',
+        './assets/css/application.scss',
+      ],
+    }
+    
+    Glob.sync("./assets/*/*.*").reduce((_, entry) => {
```
stanislas-m: If you use reduce here, you'll drop the first file, even if it's not application.scss. :)

paganotoni:
@stanislas-m all set with that change.

PR #1011, opened 2018-04-07
Fix #886: ensure generated CSRF is not overriden

### LINKED ISSUE #886: CSRF error when Clearing the Session and Redirecting to a form
We discussed the case with @markbates on Slack.

Here is the error:
```
CSRF token not found in request
github.com/gobuffalo/buffalo.sessionSaver.func1
    /home/kteb/gowork/src/github.com/gobuffalo/buffalo/session.go:72
github.com/gobuffalo/buffalo.RequestLoggerFunc.func1
    /home/kteb/gowork/src/github.com/gobuffalo/buffalo/request_logger.go:51
github.com/gobuffalo/buffalo.(*App).PanicHandler.func1
    /home/kteb/gowork/src/github.com/gobuffalo/buffalo/errors.go:69
github.com/gobuffalo/buffalo.RouteInfo.ServeHTTP
    /home/kteb/gowork/src/github.com/gobuffalo/buffalo/handler.go:75
github.com/gobuffalo/buffalo.(*RouteInfo).ServeHTTP
    <autogenerated>:1
```

Here is the code I used 
```
func SetCurrentUser(next buffalo.Handler) buffalo.Handler {
    return func(c buffalo.Context) error {
        if uid := c.Session().Get("current_user_id"); uid != nil {
            u := &models.User{}
            tx := c.Value("tx").(*pop.Connection)
            err := tx.Find(u, uid)
            if err != nil {
                if errors.Cause(err) != sql.ErrNoRows {
                    return errors.WithStack(err)
                }
                c.Session().Clear()
                c.Flash().Add("warning", "We didn't find the session you were connected to, so please reconnect")
                return c.Redirect(302, "/signin")
            }
            c.Set("current_user", u)
        }
        return next(c)
    }
}
```

I'm using the default form generated by auth:
```
<%= form_for(user, {action: signinPath()}) { %>
  <%= f.InputTag("Email") %>
  <%= f.InputTag("Password", {type: "password"}) %>
  <button class="btn btn-success">Sign In!</button>
<% } %>
```

I reproduce it with migrating down the db and migrating up and seeding my default user.

From here it's the explanation of Mark on Slack:

i can reproduce, not sure i know what the fix is, but here’s what’s happening:
* you make a call to clear the session
* the redirect happens
* buffalo renders the template, with the CSRF token in it (it’s there, i saw it in the generated HTML)
* the page is rendered, and the request to delete the session cookie completes

this is where things get interesting.
the page is already rendered with a token, *however* that token was salted using a salt that was in the session that was clear *after* the page was drawn; cause cookies.
and that’s how you get a CSRF token in a form but not have it be valid. :slightly_smiling_face:
the error message is misleading though.
it was found, but it’s not a valid CSRF token.

the work around would be to just delete the `current_user_id` from the session, but it would be great if someone can figure out a good way to stop this from happening in the future.

stanislas-m:
* In the case realToken was regenerated, a scoped redefinition was overriding
    the token to nil.

### DIFF
--- a/middleware/csrf/csrf.go
+++ b/middleware/csrf/csrf.go
@@ -65,12 +65,13 @@ var New = func(next buffalo.Handler) buffalo.Handler {
 		}
 
 		var realToken []byte
+		var err error
 		rawRealToken := c.Session().Get(tokenKey)
 
 		if rawRealToken == nil || len(rawRealToken.([]byte)) != tokenLength {
 			// If the token is missing, or the length if the token is wrong,
 			// generate a new token.
-			realToken, err := generateRandomBytes(tokenLength)
+			realToken, err = generateRandomBytes(tokenLength)
 			if err != nil {
 				return err
 			}
@@ -171,8 +172,9 @@ func compareTokens(a, b []byte) bool {
 // one-time-pad used to mask it.
 func xorToken(a, b []byte) []byte {
 	n := len(a)
-	if len(b) < n {
-		n = len(b)
+	bn := len(b)
+	if bn < n {
+		n = bn
 	}
 
 	res := make([]byte, n)


markbates:
Feel free and merge once all the tests pass. Thanks!

PR #1019, opened 2018-04-10
fix resource name handling bug

sio4:
Hi, while I generate resource named `resources`, I found a bug on name detection routine `Resource()` function.

If I generate my resource with "resource" (`buffalo generate resource resources`), the structure name of generated resource is `ResourcesResource` and `strings.Replace()` on `Resource()` of `router.go` replaces first `Resource` and remaining `sResource` is used as resource name.

I replaced `Replace()` with `TrimSuffix()` and I think it is fit for the logic too.
I also added some test cases for this.


sio4:
Thank you all for this lovely framework! @markbates @paganotoni

PR #1020, opened 2018-04-10
allow for redirects in actions using a path helper

markbates:
```go
return c.Redirect(302, "widgetPath()", render.Data{"widget_id": widget.ID})
// or
return c.Redirect(302, "widgetPath()", map[string]interface{}{"widget_id": widget.ID})
```

Replaces PR https://github.com/gobuffalo/buffalo/pull/622

### DIFF
--- a/default_context.go
+++ b/default_context.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"reflect"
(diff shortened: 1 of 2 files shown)


paganotoni:
This may be too late but what about something like:
```
return c.Redirect(302, "widgetPath", widget.ID)

PR #1025, opened 2018-04-13
updated the integration tests to use gometalinter and setup a bunch of linters that need to pass before a PR can merge.

markbates:


markbates:
@paganotoni yeah, it wouldn't be fair of me to impose this without first making sure it ran clean. :)

app.go
```diff
@@ -73,32 +70,28 @@ func (a *App) Serve() error {
...
-		err = server.Serve(listener)
-	} else {
-		// Use a TCP socket
-		server.Addr = a.Options.Addr
-
-		// start the web server
-		err = server.ListenAndServe()
+		if err = server.Serve(listener); err != nil {
+			return a.Stop(err)
+		}
```
stanislas-m: There is a missing `return nil` here, isn't it?

markbates: Maybe. ;)

PR #1027, opened 2018-04-14
Adds live-reload with LiveReloadPlugin

paganotoni:
This PR adds live CSS and JS reload by using a different strategy than previous, submitting it for your consideration.

### DIFF
--- a/generators/newapp/templates/templates/application.html.tmpl
+++ b/generators/newapp/templates/templates/application.html.tmpl
@@ -24,5 +24,10 @@
     </div>
 
     <%= javascriptTag("application.js") %>
+
+    <%= if( env == "development") { %>
+    <!-- For buffalo to do live js/css reload -->
+    <script src="http://localhost:35729/livereload.js"></script>
+    <% } %>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


generators/newapp/templates/templates/application.html.tmpl
```diff
@@ -24,5 +24,10 @@
     </div>
 
     <%= javascriptTag("application.js") %>
+
+    <%= if( env == "development") { %>
+    <!-- For buffalo to do live js/css reload -->
+    <script src="http://localhost:35729/livereload.js"></script>
```
markbates: What if we have `javascriptTag` automatically generate `<script src="http://localhost:35729/livereload.js"> </script>` too?

markbates:
@paganotoni @stanislas-m I know not of the magical ways for live reloading. :) is there a way to get this to trigger if a template has changed too?

generators/newapp/templates/templates/application.html.tmpl
paganotoni: it seems the plugin provides a way to inject this script in the header without putting it here, so i removed this portion of the changes :)

paganotoni:
I don't see an option on the plugin for the HTML, but i will continue looking for alternatives :)

paganotoni:
@markbates i'll add HTML reload support in another PR, I have a solution for it but I don't feel comfortable with it, I would like to take a bit more time reviewing alternatives.

PR #1028, opened 2018-04-14
adding some filetests to check the buffalo-version output

paganotoni:
@markbates adding a couple Filetests, when you ask for version on the vendored buffalo it will print something like v0.11.0 while on the non-vendored will say just development.

### DIFF
--- a/buffalo/cmd/filetests/version-dep.json
+++ b/buffalo/cmd/filetests/version-dep.json
@@ -1,6 +1,6 @@
 [{
   "path": "output.txt",
   "contains": [
-    "v0.11.0 [development]"
+    "msg=\"Buffalo version is: v\n"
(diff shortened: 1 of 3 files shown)


buffalo/cmd/filetests/version-dep.json
```diff
@@ -0,0 +1,6 @@
+[{
+  "path": "output.txt",
+  "contains": [
+    "v0.11.0 [development]"
```
markbates: This will fail every time we have a new release. Got anything else you can use here instead?

paganotoni:
@markbates changed it a bit :)

markbates:
@paganotoni looks good to me. feel free to merge when the tests pass. thanks.

PR #1037, opened 2018-04-20
Make app version string available from inside app (v2) fixes #1035

### LINKED ISSUE #1035: Make app version string available from inside app
(From slack chat)

I'd like to return my app version as a response header, but it's not currently available in any simple way. It'd be great to expose this somehow.

Thanks!

image

markbates:


markbates:
@stanislas-m good catch! fixed.

PR #1054, opened 2018-05-08
moves to use bootstrap 4.1.1 to avoid a text-hide warning 

paganotoni:
Latest bootstrap releases removes a warning when compiling with the text-hide mixin, I've seen that warning in our development version since we move to use bootstrap 4.1.0 as default.

This PR moves buffalo to use 4.1.1 instead of 4.1.0.


paganotoni:
@stanislas-m do you by chance know what should I do in appveyor to make it pass?

stanislas-m:
@paganotoni #1050 will fix the problem. :)

PR #1055, opened 2018-05-08
Task/adding resource fields validation

paganotoni:
Adds validation to resource fields names as mentioned in [placeholder] ticket.

### DIFF
--- a/generators/resource/generator.go
+++ b/generators/resource/generator.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"errors"
+	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


markbates:
@paganotoni I like what you're going for here, but what if I was to use `buffalo db g model`? then I wouldn't have the field validation. Perhaps this should be in Pop and then called from the resource generator via a package. Thoughts?

paganotoni:
@markbates agree with you, I think there is a PR which is WIP, that one will cover this as well, thanks for the guidance on this PR.

PR #1071, opened 2018-05-22
Force a default session secret in development environment

dnnrly:


options.go
```diff
@@ -131,8 +131,12 @@ func optionsWithDefaults(opts Options) Options {
 	if opts.SessionStore == nil {
 		secret := envy.Get("SESSION_SECRET", "")
 		// In production a SESSION_SECRET must be set!
-		if opts.Env == "production" && secret == "" {
-			logrus.Warn("Unless you set SESSION_SECRET env variable, your session storage is not protected!")
+		if secret == "" {
+			if opts.Env == "production" {
```
markbates: This is too loose. The logic should be if dev or test then set else error. Only test and dev get a free pass. Everyone else has to setup the var. it’s a safety issue.

markbates:
closing in favor of #1083

PR #1094, opened 2018-06-09
Allowing to skip schema load

paganotoni:
In some situations, the database user configured in the database connection is not the owner of the plpgsql extension, and Postgres ends up saying when we try to load the schema file.

```
Error: unable to load schema for application_test: pq: must be owner of extension plpgsql
```

This PR allows to skip schema load and go straight to migrations.

### DIFF
--- a/buffalo/cmd/test.go
+++ b/buffalo/cmd/test.go
@@ -22,9 +22,12 @@ import (
 const vendorPattern = "/vendor/"
 
 var vendorRegex = regexp.MustCompile(vendorPattern)
+var skipSchemaLoad = false
 
 func init() {
 	decorate("test", testCmd)
+	testCmd.Flags().BoolVarP(&buildOptions.ExtractAssets, "skip-schema-load", "s", false, "skips loading the schema and instead runs migrations before running tests")
+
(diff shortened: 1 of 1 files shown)


buffalo/cmd/test.go
```diff
@@ -22,9 +22,12 @@ import (
 const vendorPattern = "/vendor/"
 
 var vendorRegex = regexp.MustCompile(vendorPattern)
+var skipSchemaLoad = false
 
 func init() {
 	decorate("test", testCmd)
+	testCmd.Flags().BoolVarP(&buildOptions.ExtractAssets, "skip-schema-load", "s", false, "skips loading the schema and instead runs migrations before running tests")
```
markbates: I'm not a fan of the name of this flag. Any other options?

paganotoni: Oh boy! ouch! going for it.

paganotoni: @markbates  what about "forceMigrations" and -m --force-migrations ?

paganotoni:
@markbates moved the variable to be `forceMigrations` (-m and --force-migrations in the cli).

PR #1099, opened 2018-06-12
Make the flag --config global in all buffalo commands

### LINKED ISSUE #211: Allow using a "default" config for generating new Buffalo projects
At the bottom of https://github.com/gobuffalo/buffalo/blob/master/buffalo%2Fcmd%2Froot.go is a bunch of commented out code that cobra generated to read in a `$HOME/.buffalo.yml` file so people can set up their favorite buffalo configs for generating new projects, like their preferred database. It would be great to actually support that.

isgj:
Fixes #211 At the moment only the *buffalo new* sub-command make use of the global config. The order for evaluating  a flag is 1- default value, 2 - value read from the config file, 3 - the value from the cmd line. So if the .buffalo.yaml 
```yaml
api: true
db-type: mysql
```
and you start a new project with `buffalo new app_name --db-type=postgres`
the new project will be an _'api'_ because `api: true` was read from the config file but the db it will use is `postgres` because the cmd line value will override the flag.


isgj:
After looking at the **`db`** sub-command(pop/soda/) this change will override the `--config` flag of `buffalo db`. Will it be better to make the global `--config` flag local to the `new` cmd?

markbates:
Yeah, it should be scoped to “buffalo new” for now.

robbyoconnor:
Pay attention to your branch ;)

isgj:
@robbyoconnor next time will branch from development

robbyoconnor:
You should probably close it and do it against that branch ;)

PR #1100, opened 2018-06-12
Add codecov.io badge

AlekSi:
There is no coverage information for `master` branch yet, that's why badge says "unknown" for now.

`.codecov.yml` format is documented [here](https://docs.codecov.io/docs/codecov-yaml). See also next pages; specifically, [here](https://docs.codecov.io/docs/coverage-configuration) for what "range" means.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,2 @@
+coverage:
+  range: "60...70"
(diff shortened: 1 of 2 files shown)


AlekSi:
btw, @markbates, you can add `--squash` to `docker build` for `gobuffalo/buffalo:development` image to make it significantly smaller.

markbates:
@AlekSi if you know how to do that on hub.docker.com please let me know. That's where we build the images.

AlekSi:
Looks like it is not possible: https://github.com/docker/hub-feedback/issues/955 One workaround is to use multi-stage builds which are supported. Another one is to rebuild this image somewhere else.

markbates:
@AlekSi feel free and take a crack at making it smaller.

PR #1119, opened 2018-06-19
Make the --config flag of *buffalo new* command use the config file

### LINKED ISSUE #211: Allow using a "default" config for generating new Buffalo projects

isgj:
Fixes #211 As suggested on thread #1099 branched from development directly.


buffalo/cmd/new.go
```diff
@@ -121,24 +136,38 @@ func notInGoPath(ag newapp.Generator) error {
...
+	newCmd.Flags().Bool("skip-webpack", false, "skips adding Webpack to your app")
+	newCmd.Flags().Bool("skip-yarn", false, "use npm instead of yarn for frontend dependencies management")
+	newCmd.Flags().String("db-type", "postgres", fmt.Sprintf("specify the type of database you want to use [%s]", strings.Join(pop.AvailableDialects, ", ")))
+	newCmd.Flags().String("docker", "multi", "specify the type of Docker file to generate [none, multi, standard]")
+	newCmd.Flags().String("ci-provider", "none", "specify the type of ci file you would like buffalo to generate [none, travis, gitlab-ci]")
+	newCmd.Flags().String("vcs", "git", "specify the Version control system you would like to use [none, git, bzr]")
+	newCmd.Flags().Int("bootstrap", 3, "specify version for Bootstrap [3, 4]")
+	viper.BindPFlags(newCmd.Flags())
+	cobra.OnInitialize(initConfig)
+	newCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.buffalo.yaml)")
```
robbyoconnor: I dislike this -- why not add a sane set of defaults and place the config in the buffalo project dir?

buffalo/cmd/new.go
```diff
@@ -121,24 +136,38 @@ func notInGoPath(ag newapp.Generator) error {
...
+		viper.SetConfigName(".buffalo") // name of config file (without extension)
+		viper.AddConfigPath("$HOME")    // adding home directory as first search path
+		viper.AutomaticEnv()            // read in environment variables that match
+	}
+
+	// If a config file is found, read it in.
+	if err := viper.ReadInConfig(); err == nil {
+		fmt.Println("Using config file:", viper.ConfigFileUsed())
+	}
 }
```
robbyoconnor: There is no need to init the config file if we don't provide one -- use sane defaults in-memory but toss it in the buffalo project dir, not the home dir if this is a non-negotiable.

buffalo/cmd/new.go
pkgmain: This is a cobra convention:
https://github.com/spf13/cobra/blob/master/cobra/cmd/init.go#L183

robbyoconnor:
Is that not optional at all?

buffalo/cmd/new.go
robbyoconnor: I suppose maybe we can use `$HOME/.buffalo/.buffalo.yml`??

isgj: @robbyoconnor I used `$HOME/.buffalo.yaml ` because it was the default generated but it can be changed easily. If I'm not wrong @markbates mentioned it to an issue to be there.

markbates: @robbyoconnor this is used with the buffalo new command, so it can’t live in the app, and needs to live somewhere “common”


Cobra, by convention, looks in the home directory for those dot files. Rails does, as well as a lot of similar projects, so I think that follows the path of least surprise. 

A loose feature “spec”:
By default buffalo new should read that file and use its defaults, if any. 

If no file, it should use the defaults in the bin. 

There should be a flag that lets you select a different config file other than the one at home. 

If the user passes a flag, it overrides anything in the config or in the bin 

Finally there should be a way of disabling the config file if the user doesn’t want to use it for a specific app

isgj: The first 4 "specs" are "respected". As for the last maybe not so obvious you can pass a non existing file or a file of a different format that cobra can't read to the `--config` flag

markbates: if someone passes a non-existent or badly formatted file it should error, not ignore the workflow. Why can't we add a --skip-config bool flag?

robbyoconnor:
Rebase :)

buffalo/cmd/new.go
robbyoconnor: Then I'll bow out here.

markbates:
@isgj would you mind adding something to the README.md with an example of what the config file would look?

markbates:
Yay! Love it. Been wanting to add this since the beginning. Thank you.

Issue #1116, opened 2018-06-18
Proposal: Support for yarn builds


dnnrly:
As a developer
I would like to build my frontend assets with yarn
So that I can use the same workflow I familiar with, reusing code from other codebases

<hr />

There are only 3 sub-commands that need to worry about yarn. `build` and `dev` obviously but also `test` should be considered.

Using yarn for building assets means defining entries under the `scripts` object. Let's just call them:

* `buffalo:build` -> `webpack -p --progress`
* `buffalo:dev` -> `webpack --watch`
* `buffalo:test` - will have to think about this one

These would be inserted in to the `package.json` template but only if Yarn is enabled. We could just assume that if you want to perform a yarn build, you must support those scripts.

Adding scripts to `package.json` means that upgrades will require a change to the template of that file. Unless we are somehow able to parse and dynamically add those scripts to the file, we'll have to use the `upgrade` sub command to overwrite it with the new version.

For plumbing in the commands that Buffalo will invoke, we know whether yarn is installed, running `yarn run` will display the available scripts. We can parse this output for the relevant entries.
For `build`, we need to modify `./cmd/build.buildAssets` to test for `buffalo:build` if we have yarn.
For `dev`, we need to modify `./cmd.startWebpack` for `buffalo:dev`.
The tricky one is `test` as this doesn't currently support testing assets. To achieve this we would need to use a similar approach to `dev` so checking for the presence of yarn and it's scripts. If they are there, we can run the command and test it's output.

stanislas-m:
Fixed with #1718

PR #1136, opened 2018-06-23
Yarn based asset building

dnnrly:
This hopes to solve #1116


dnnrly:
It was MUCH simpler than I anticipated! But I've made a few opinionated decisions - I fully expect (nae, insist) that the community has some input.

robbyoconnor:
:100:

mclark4386:
I love this! it makes it so that you can that much more easily customize your asset build pipeline! Hope this gets merged soon!

dnnrly:
That's what I figured, I'm not in a rush. How long before the next major update?
I guess it depends what the appetite is for this change. User poll in Slack?

The question is how you determine whether those scripts are present. You could "grep" `package.json` - yuck. Or parse it, still yuck but not so much. Ask yarn? That could be slow and lead to non obvious errors if you're mid feature etc.

Not sure how you would set sticky config to just tell buffalo "You are now using yarn all the way". That would be my preference I think.

markbates:
It’ll probably be a while for the next major rev. With that said, it’s just json, couldn’t you pull it into a simple map and ask it if has those scripts? If so, use them, otherwise fall back to current behavior.

dnnrly:
For my notes when life let's me get round to it:
https://stackoverflow.com/questions/11066946/partly-json-unmarshal-into-a-map-in-go

buffalo/cmd/test.go
```diff
@@ -73,6 +73,20 @@ var testCmd = &cobra.Command{
 				}
 			}
 		}
+		if meta.New(".").WithYarn {
```
dnnrly: Can we inject an `app` in to this function instead?

buffalo/cmd/build.go
```diff
@@ -40,6 +40,14 @@ var xbuildCmd = &cobra.Command{
 			b.Tags = append(b.Tags, buildOptions.Tags)
 		}
 
+		if b.WithYarn {
+			found, err := YarnScriptExists("buffalo:dev")
+			if err != nil {
+				logrus.Fatal(err)
+			}
+			b.WithYarn = found
```
dnnrly: This feels like I'm abusing a flag that just happens to be in the right place. This 'makes it work' but should I do something else to pass this information in to the builder?

dnnrly:
@markbates it finally passes tests, including an extra integration test. But does it pass muster?

gedw99:
This is a great pr.
Thanks @dnnrly !

buffalo/cmd/dev.go
```diff
@@ -91,6 +91,13 @@ func startWebpack(ctx context.Context) error {
 	}
 
 	cmd := exec.CommandContext(ctx, webpack.BinPath, "--watch")
+	if app.WithYarn {
+		if found, err := YarnScriptExists("buffao:dev"); err != nil {
```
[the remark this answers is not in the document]
dnnrly: Dammit! It'll have to wait until this evening I'm afraid. 😢

buffalo/cmd/build/assets.go
```diff
@@ -9,12 +9,18 @@ import (
...
 
-	if b.WithWebpack && b.Options.WithAssets {
+	if b.Options.WithAssets {
 		if err := envy.MustSet("NODE_ENV", "production"); err != nil {
 			return errors.WithStack(err)
 		}
-		if err := b.exec(webpack.BinPath); err != nil {
-			return errors.WithStack(err)
+		if b.WithYarn {
+			if err := b.exec("yarn", "buffalo:build"); err != nil {
```
markbates: If there isn’t a script this will error, so it will break existing applications.

buffalo/cmd/dev.go
```diff
@@ -68,8 +68,8 @@ This behavior can be changed in your .buffalo.dev.yml file.`,
 
 func startWebpack(ctx context.Context) error {
 	app := meta.New(".")
-	if !app.WithWebpack {
-		// there's no webpack, so don't do anything
+	if !app.WithWebpack && !app.WithYarn {
+		// there's no webpack or yarn, so don't do anything
```
markbates: Do you need this check? If there’s no webpack config is there yarn? Or rather what’s the situation where there’s yarn, but no webpack?

buffalo/cmd/build.go
```diff
@@ -40,6 +40,14 @@ var xbuildCmd = &cobra.Command{
 			b.Tags = append(b.Tags, buildOptions.Tags)
 		}
 
+		if b.WithYarn {
+			found, err := YarnScriptExists("buffalo:dev")
+			if err != nil {
+				logrus.Fatal(err)
```
markbates: I don’t think we need Fatal here, it should just fall back to regular webpack

buffalo/cmd/dev.go
```diff
@@ -91,6 +91,13 @@ func startWebpack(ctx context.Context) error {
 	}
 
 	cmd := exec.CommandContext(ctx, webpack.BinPath, "--watch")
+	if app.WithYarn {
+		if found, err := YarnScriptExists("buffalo:dev"); err != nil {
+			return errors.WithStack(err)
+		} else if found {
```
markbates: I feel like this can be collapsed to remove the else if. I had to read it a few times to understand the meaning of it.

buffalo/cmd/dev.go
dnnrly: If you want to use a yarn script, you don't necessarily need to use webpack. With the template that is supplied as part of this PR we can use either webpack or yarn but it would be very easy for someone to modify the yarn script to do something else - such as building the assets directly.

PR #1137, opened 2018-06-25
LooseSlash work correctly 

### LINKED ISSUE #955: trailing slash in route stripped regardless of LooseSlash option
App with these routes:
```
app.GET("/top-comments-abc", TopCommentList)
app.GET("/top-comments-abc/", TopCommentList)
```

and this logging in router.go:

```
func (a *App) addRoute(method string, url string, h Handler) *RouteInfo {
    a.moot.Lock()
    defer a.moot.Unlock()
    l := log.New(os.Stdout, "", 0)
    l.Println("router url : " + url)

    url = path.Join(a.Prefix, url)
    name := a.buildRouteName(url)

    l.Println("router url post join : " + url)
    l.Println("router name : " + name)
```

prints this on build:

```
router url : /
router url 2 : /
router name : root
router url : /top-comments-abc
router url post join : /top-comments-abc
router name : topCommentsAbc
router url : /top-comments-abc/
router url post join : /top-comments-abc
router name : topCommentsAbc
```

which suggests that 

``` 
url = path.Join(a.Prefix, url)
```

 is removing the trailing slash.

Sometimes that trailing slash is necessary.

isgj:
Fixes #955 This PR is against development branch
#1135


isgj:
Since the the default behavior wasn't working right before probably some tests will fail. I think it should be labeled as a breaking change

markbates:
Closing in favor of #1168

PR #1138, opened 2018-06-25
Refactor CI config generator

stanislas-m:
* Add MySQL conf for Travis setup.
* Extract CI generator, so it can be reused (so we can generate the CI conf after creating the app).

### DIFF
--- a/generators/ci/ci.go
+++ b/generators/ci/ci.go
@@ -12,6 +12,7 @@ func (cg Generator) Run(root string, data makr.Data) error {
 	should := func(data makr.Data) bool {
 		return cg.Provider != "none"
 	}
+	data["opts"] = cg
 	var f makr.File
 	switch cg.Provider {
 	case "travis":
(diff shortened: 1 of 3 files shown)


markbates:
I’ll check this out later, but what are your thoughts on moving these out into plugins? Would be faster to iterate on, thin out core a bit more, and set a precedent for using plugins for stuff like this. Thoughts?

paganotoni:
Sorry to jump in here but I love the idea of moving CI things out to be plugins

stanislas-m:
That seems to be a good idea, but it means we'll remove the option from the `new` command, right?

markbates:
Eventually we have to lose the flag, but we migrate by putting out the plugins first and then updating the flag to go get and run the plugin and issue a deprecation warning saying use the plugin in the future.

markbates:
@stanislas-m I'm closing this since we both agree it should move to a plugin.

stanislas-m:
Sure, np.

PR #1141, opened 2018-06-27
Clear flashes when using redis

### LINKED ISSUE #911: Flashes won't clear when using redis-backed session storage
Seeing that buffalo uses gorilla's sessions package, I wanted to replace the cookie storage with a redis-backed storage from `gopkg.in/boj/redistore.v1`. When I did this, flash messages won't clear after dismissing them and reloading the page.

I created a repo at [lflux/sessionthing](https://github.com/lflux/sessionthing) where you can reproduce this by:
1. Setting up a Redis server at `:6379`
1. Cloning the repo and running `buffalo dev`
1. Visit http://localhost:3000/flash which will redirect to `/`
1. See the flash message `Success flash`. Dismiss and reload the page
1. The flash message will persist.

By setting `useRedis = false` in app.go which uses the default cookie storage, the flash message will no longer persist over page loads.

isgj:
In the documentation of gorilla/sessions it says that the Store should be saved before writing to the response (or before returning from the handler) so this works with CoockieStore and redis-backend.
Fixes #911

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ Please take moment and read this fantastic post on how to easily work with Go re
 
 ## Documentation Welcome
 
-Hands down the most important, and the most welcome, pull requests are for documentation. We LOVE documentation PRs, and so don't all those that come after you.
+Hands down the most important, and the most welcome, pull requests are for documentation. We LOVE documentation PRs, and so do all those that come after you.
 
 Whether it's GoDoc or prose on [http://gobuffalo.io](http://gobuffalo.io) all documentation is welcome.
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ ADD . .
 RUN go get -v -t ./...
 RUN make install
 
-RUN go test -tags sqlite -race ./...
+RUN go test -tags sqlite -race  ./...
 RUN go test -tags sqlite -coverprofile cover.out -covermode count ./...
 
 RUN if [ -z "$CODECOV_TOKEN"  ] ; then \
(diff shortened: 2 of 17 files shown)


isgj:
Actually the session is saved

markbates:
Where?

isgj:
Render when you the flashes are cleared or at the beginning of the Redirect method. Before the Save on session was call more then once.

markbates:
What about if Redirect and Render aren’t called? That middleware caught all possible exits. This doesn’t.

isgj:
You are right. As soon as i get back to the pc i will add back the middleware

markbates:
No worries. That’s why we have PRs. :)

isgj:
I ended up to changed only one line. The problem was that `defer` called the function with the values it found before `a.Middleware.handler....` was called. Also 
```go
defer func (){ c.Flash().persist(c.Session()) }()
```
should have work, but I save the session only on error. If there are no errors `sessionSaver` has saved the session.

paganotoni:
This one is great! thanks @isgj !

jaredfolkins:
I wanted to leave a note here that this is still broken as of v0.13.12.

The filesystem store also suffers from this bug.

The fix is almost the same but it appears that buffalo has gone through a large refactor. The code now lives in `buffalo/route_info.go`.

I ran `go mod init` and `go mod vendor` to pull down a local copy of the project. I then refactored ServeHTTP as the PR from @isgj had stated.

```
func (ri RouteInfo) ServeHTTP(res http.ResponseWriter, req *http.Request) {
	defer gcontext.Clear(req)

	a := ri.App

	c := a.newContext(ri, res, req)


	payload := events.Payload{
		"route":   ri,
		"app":     a,
		"context": c,
	}

	events.EmitPayload(EvtRouteStarted, payload)

	err := a.Middleware.handler(ri)(c)

	if err != nil {
                c.Flash().persist(c.Session())
		events.EmitError(EvtRouteErr, err, payload)
		// things have really hit the fan if we're here!!
		a.Logger.Error(err)
		c.Response().WriteHeader(500)
		c.Response().Write([]byte(err.Error()))
	}
	events.EmitPayload(EvtRouteFinished, payload)
}
```

I then simply ran `go run -mod=vendor main.go` from my project root. 

Obviously less than ideal but for now it'll work.

Thanks for all the work Mark & Contributors!

PR #1149, opened 2018-06-29
automatically respond with a 404 if error is sql.ErrNoRows

markbates:


markbates:
Sorry for all of the moving files around, but drove me nuts when trying to write the tests.

PR #1172, opened 2018-07-15
Ensure no information lost in binding time format

### LINKED ISSUE #1169: One of Time Formats in binding is Illogical
The following registered format doesn't make sense to have it:
https://github.com/gobuffalo/buffalo/blob/25619606ab2ac525ee38a07a15bfcffc832125b4/binding/binding.go#L33

It loses the information of which half of the day it's part of (AM or PM). Either the format `2006-01-02T15:04` should be added to the list, or the format is replaced with `2006-01-02T15:04`.

I'll get back to fix this once I'm done with the project at hand. Meanwhile, which solution is preferred?

Mohammed90:
One of the currently registered time formats loses whether the parsed time is in AM or PM. Furthermore, it could fail if the input time is within the hours 13-to-00.

Fixes #1169

### DIFF
--- a/binding/binding.go
+++ b/binding/binding.go
@@ -30,7 +30,7 @@ var timeFormats = []string{
 	"2006-01-02T15:04:05Z07:00",
 	"01/02/2006",
 	"2006-01-02",
-	"2006-01-02T03:04",
+	"2006-01-02T15:04",
 	time.ANSIC,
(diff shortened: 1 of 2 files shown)


Mohammed90:
Done!

paganotoni:
thanks @Mohammed90 !

PR #1175, opened 2018-07-16
feature: download renderer to facilitate file downloads

pkgmain:
This is what I'm using for file downloads in my projects so I thought I'd open a PR to see if there's any desire for it in Buffalo proper. Example usage:

```go
c.Render(http.StatusOK, r.Download(c, "report.pdf", bytes.NewReader(pdfBytes))
```

### DIFF
--- a/render/download.go
+++ b/render/download.go
@@ -1,0 +1,62 @@
+package render
+
+import (
+	"fmt"
+	"io"
+	"mime"
+	"net/http"
+	"path/filepath"
+	"strconv"
+)
+
+type downloadRenderer struct {
+	data     []byte
+	filename string
+	writer   http.ResponseWriter
+}
+
+func (r downloadRenderer) ContentType() string {
+	ext := filepath.Ext(r.filename)
+	t := mime.TypeByExtension(ext)
+	if t == "" {
+		t = "application/octet-stream"
+	}
+
+	return t
+}
+
+func (r downloadRenderer) Render(w io.Writer, d Data) error {
+	r.writer.Header().Add("Content-Disposition", fmt.Sprintf("attachment; filename=%s", r.filename))
+	r.writer.Header().Add("Content-Length", strconv.Itoa(len(r.data)))
+
+	_, err := w.Write(r.data)
+	return err
+}
+
+// Download renders a file attachment automatically setting following headers:
+//
+//   Content-Type
+//   Content-Length
+//   Content-Disposition
+//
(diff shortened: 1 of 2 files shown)


render/download.go
```diff
@@ -0,0 +1,62 @@
...
+	t := mime.TypeByExtension(ext)
+	if t == "" {
+		t = "application/octet-stream"
+	}
+
+	return t
+}
+
+func (r downloadRenderer) Render(w io.Writer, d Data) error {
+	r.writer.Header().Add("Content-Disposition", fmt.Sprintf("attachment; filename=%s", r.filename))
```
markbates: Cast w if it’s an http writer,or better yet it’s a writer that implements an interface that has Header() then write the headers.

render/download.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Download renders a file attachment automatically setting following headers:
+//
+//   Content-Type
+//   Content-Length
+//   Content-Disposition
+//
+// Content-Type is set using mime#TypeByExtension with the filename's extension. Content-Type will default to
+// application/octet-stream if using a filename with an unknown extension.
+func Download(data []byte, filename string, writer http.ResponseWriter) Renderer {
```
markbates: Args should be name string and r Reader

render/download.go
markbates: Yes, sorry, you're right. That's what happens when you do PR reviews before coffee. :)

render/download.go
markbates: In that case you should change the signature model to look more like: https://github.com/gobuffalo/buffalo/blob/master/render/auto.go#L43

```go
func Download(ctx buffalo.Context, name string, r io.Reader) Renderer
```

markbates: then you have everything from the context, so you can do whatever you want and you can use anything for `io.Reader`.

pkgmain: I think using a `buffalo.Context` would cause a cyclical import, no?

markbates: d'oh, I meant `context.Context`, like here:
https://github.com/gobuffalo/buffalo/blob/master/render/auto.go#L43

Then you can check against a private interface:

```go
type responsible interface {
  Response() http.ResponseWriter
}

func Download(ctx context.Context, name string, r io.Reader) Renderer {
  bc, ok := ctx.(responsible)
  if ok {
// ...
  }
// ...
}
```

Something like that anyway. That way no cyclical imports, easily testable, etc... hopefully.

pkgmain: Gotcha, that makes sense. Updated.

PR #1176, opened 2018-07-16
make sure all caught panics have a stack with them

markbates:


errors.go
```diff
@@ -58,12 +58,12 @@ func (a *App) PanicHandler(next Handler) Handler {
 			if r != nil { //catch
 				switch t := r.(type) {
 				case error:
```
marwan-at-work: you might have to do `err = t` otherwise err is nill still

markbates: good catch!

markbates:
@marwan-at-work fixed! and added some tests to make sure. thanks for the review!

PR #1186, opened 2018-07-19
Remove logger need from the render package

stanislas-m:
If a rendering engine is not found for a given extension, it should just return an error.

### DIFF
--- a/render/template.go
+++ b/render/template.go
@@ -1,6 +1,7 @@
 package render
 
 import (
+	"fmt"
 	"html/template"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


stanislas-m:
Yeah, I understand this is a breaking change. My point is, as a buffalo user, silent warnings like that can be easily missed. If I use a specific extension, I'm expecting it to work or fail (and I think most users will use only one extension).

On the testing side, I do think having an error here instead of a warning can help to find mistakes.

render/template.go
```diff
@@ -119,8 +119,7 @@ func (s templateRenderer) exec(name string, data Data) (template.HTML, error) {
 	for _, ext := range s.exts(name) {
 		te, ok := s.TemplateEngines[ext]
 		if !ok {
-			logrus.Errorf("could not find a template engine for %s\n", ext)
-			continue
+			return "", fmt.Errorf("could not find a template engine for %s", ext)
```
robbyoconnor: Couldn't this use the configured logger?

PR #1187, opened 2018-07-19
Generify logrus usage

stanislas-m:
Remove logrus references where it's not needed, use the logger configured in Options if possible.

### DIFF
--- a/options.go
+++ b/options.go
@@ -6,8 +6,6 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/fatih/color"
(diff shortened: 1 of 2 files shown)


options.go
```diff
@@ -131,7 +129,7 @@ func optionsWithDefaults(opts Options) Options {
 			if opts.Env == "development" || opts.Env == "test" {
 				secret = "buffalo-secret"
 			} else {
-				logrus.Warn("Unless you set SESSION_SECRET env variable, your session storage is not protected!")
+				opts.Logger.Warn("Unless you set SESSION_SECRET env variable, your session storage is not protected!")
```
markbates: Could the logger be nil here?

PR #1192, opened 2018-07-23
renamed "buffalo update" to "buffalo fix" to match "go fix".

markbates:
I kept the `update` alias to make it backward compatible.


markbates:
Good catch!

--
Mark Bates

Issue #856, opened 2018-01-13
Allow cookies to have a custom path


stanislas-m:
Cookies are bound on a path, and by default it's the current "directory" in the URL: for instance, if I set a cookie in a "/users/set-cookie" URL, the cookie will be bound to "/users".

We should allow Buffalo to set cookies on a custom path, so I can set a site-wide ("/") cookie from a "/users/set-cookie" path.

orderbynull:
@stanislas-m is it still relevant? I'd like to make contribution.

stanislas-m:
@orderbynull It is, go ahead! :+1:

stanislas-m:
Fixed with #1196.

PR #1196, opened 2018-07-24
ISSUE-856 Added SetWithPath cookie method

orderbynull:
This PR refers issue https://github.com/gobuffalo/buffalo/issues/856


cookies_test.go
```diff
@@ -40,6 +40,18 @@ func TestCookies_Set(t *testing.T) {
...
+	res := httptest.NewRecorder()
+
+	c := Cookies{&http.Request{}, res}
+
+	c.SetWithPath("name", "Rob Pike", "/foo")
+
+	h := res.Header().Get("Set-Cookie")
+	r.Equal("name=\"Rob Pike\"; Path=/foo", h)
+}
+
```
robbyoconnor: Shouldn't you also test with the "/" path to ensure that this cookie is not set for that?

robbyoconnor: Shouldn't you also test with the `/` path to ensure that this cookie is not set for that?

orderbynull: Not sure what do you mean. If i set c.SetWithPath for path '/' it will set.

markbates:
Thanks!

PR #1203, opened 2018-07-30
Add post initializers support

stanislas-m:
PostInitializers introduces a new way to handle a common pattern:
when you want to init some stuff or start a service just after the app
started successfully, you usually just drop the relevant code after the
routes init, in the `if app == nil {` block.

This way to do is just a workaround, and doing so can result in multiple
init, if you use `App()` more than once.

PostInitializers proposes a cleaner solution for this pattern. It allows
the user to define a slice of handlers to execute after the app starts.

### DIFF
--- a/options.go
+++ b/options.go
@@ -52,16 +52,21 @@ type Options struct {
 	// PreHandlers are http.Handlers that are called between the http.Server
 	// and the buffalo Application.
 	PreHandlers []http.Handler
-	// PreWare takes an http.Handler and returns and http.Handler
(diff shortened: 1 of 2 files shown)


stanislas-m:
Typical usage: https://github.com/gobuffalo/gobuffalo/blob/master/actions/app.go#L108

markbates:
I like the idea, but I’ve always envisioned it in my head as Events. You could register handlers for application lifecycle events like start, stop, etc...

stanislas-m:
@markbates Makes sense, yeah.

PR #1204, opened 2018-08-02
Filtered param logger

paganotoni:
This PR makes the param logger consider a blacklist for parameters content and replace these with [FILTERED].


paganotoni:
@mclark4386 can you recheck it ? need your approval again :)

PR #1205, opened 2018-08-02
Time format that allows to parse PM time as well

haraldLmueller:
without that each PM time (24h format) will be silently ignored


paganotoni:
@haraldLmueller thanks for this PR, i we merged this fix to development in https://github.com/gobuffalo/buffalo/pull/1172

Issue #1105, opened 2018-06-15
Proposal: Move Middleware to Separate Repos


markbates:
Currently there are a handful of middlewares in the "core" repo. I propose that we move each middleware to it's one repository. For example: github.com/gobuffalo/mw-poptx or github.com/gobuffalo/mw-tokenauth, etc...

The benefits to this are:

* Improved development speed of middleware 
* Reduced dependencies. For example, if you aren't using the PopTransaction middleware you still have a dependency on Pop since the middleware is part of the core repo.
* Allows for more middleware packages that might now have been part of the "core" repo.

My proposal for making these changes is:

1. Copy the current packages into new repositories
1. Deprecate the existing middleware in `v0.12.1`
1. Update `buffalo update` to write the import paths for the plugins to the new repos.
1. Remove the middleware in `v0.13.0`

robbyoconnor:
I would like to claim step 3

robbyoconnor:
Shouldn't steps 4 and 5 be swapped?

stanislas-m:
There is not particular order for these tasks. :)

robbyoconnor:
I assumed there was

PR #1213, opened 2018-08-06
Update new project template to use extracted middlewares

stanislas-m:
This is step 3 of #1105.


robbyoconnor:
I assume the updater will migrate existing projects?

stanislas-m:
@robbyoconnor Yeah, it's step 5 of #1105.

PR #1216, opened 2018-08-06
removed the weird blackfriday dep

markbates:


stanislas-m:
@markbates Is this one ready for review?

markbates:
Yep! Unless it has WIP I assume it’s ready for review. But maybe we should formalize more. Move that discussion to slack core.

PR #1219, opened 2018-08-07
Add nominal new app integration test

stanislas-m:
Here is what a nominal test case for `buffalo new coke` looks like.


markbates:
I can’t wait until we switch everything to genny. It’s crazy easy to test!

stanislas-m:
@markbates In integration tests we attack the CLI interface just at the same level as the current Docker tests. genny will help, but not for this one. :)

PR #1222, opened 2018-08-08
Update inline doc for pop transaction

showcheap:
Based on https://github.com/gobuffalo/buffalo/blob/master/middleware/pop_transaction.go#L23-L30
`middleware.PopTransaction(models.DB)` set pop.Connection


stanislas-m:
Nice catch! Thanks!

PR #1226, opened 2018-08-13
make go.mod file work correctly

rogpeppe:
The go.mod file did not make buffalo installation that would compile.
I created this one by doing a `go get github.com/gobuffalo/buffalo/...`,
and using the resulting dependency versions, so now you should
get exactly those dependencies when using Go 1.11 modules.


markbates:
Dup of #1224. Thanks.

PR #1228, opened 2018-08-14
removes the broken black friday dep and improves go modules support

markbates:


markbates:
@stanislas-m yep, just noticed that too

markbates:
@stanislas-m fixed, can you review again?

PR #1230, opened 2018-08-15
run go mod tidy

rogpeppe:
This deletes unused dependencies from go.mod and go.sum and ensures all deps for
all tags are included.


markbates:
Thanks for the heads up on `mod tidy` I'm using it now. :)

Issue #1181, opened 2018-07-18
Buffalo dev with ADDR keeps showing 127.0.0.1


jtorvald:
## Steps to Reproduce the Problem

When I use .env to set ADDR or I run buffalo dev with ADDR=0.0.0.0 buffalo dev I keep seeing:  level=info msg="Starting application at http://127.0.0.1:3000"
After trying I figured out that it actually works and it is just a wrong debug message. So this is a minor issue.

## Expected Behavior

If I config: ADDR=0.0.0.0 I expect to see
level=info msg="Starting application at http://0.0.0.0:3000"

## Actual Behavior

If I config: ADDR=0.0.0.0 I see
level=info msg="Starting application at http://127.0.0.1:3000"

When I change the PORT=3001 it works, but not for the address

## Info

Buffalo version is: v0.12.2

stanislas-m:
Fixed with #1183.

PR #1238, opened 2018-08-21
Use the right env var to change listening IP

leodotcloud:
The generated file uses `ADDR` instead of the correct variable `HOST`


robbyoconnor:
This needs to be made against `development`.

stanislas-m:
From https://godoc.org/github.com/gobuffalo/buffalo#Options:
```go
// Addr is the bind address provided to http.Server. Default is "127.0.0.1:3000"
// Can be set using ENV vars "ADDR" and "PORT".
Addr string
// Host that this application will be available at. Default is "http://127.0.0.1:[$PORT|3000]".
Host string
```

`ADDR` is the correct one, it's the TCP port used by buffalo to bind. Host is the public address, this is different.

stanislas-m:
The problem you were trying to fix was already fixed in `development`: https://github.com/gobuffalo/buffalo/issues/1181

PR #1242, opened 2018-08-22
stops buffalo new from generating incomplete projects if the templates are missing

markbates:


generators/newapp/generator.go
```diff
@@ -56,8 +60,29 @@ func New(name string) (Generator, error) {
 	return g, g.Validate()
 }
 
+const header = "Your `buffalo` binary installation is corrupted and is missing vital templates for app creation.\n"
+const footer = "Please recheck your installation: https://gobuffalo.io/en/docs/installation."
+const goPathAbuse = `It appears you are using multiple GOPATHs:
+%s
+
+Using multiple GOPATHs can cause issues with many third party tooling. Please try using only GOPATH.
```
stanislas-m: > Please try using only GOPATH

Only **one** GOPATH, maybe?

PR #1246, opened 2018-08-23
fixes an issue with go.mod not being there when running go get

markbates:


generators/newapp/templates/go.mod.tmpl
```diff
@@ -0,0 +1 @@
+module {{.opts.App.PackagePkg}}
```
camuthig: I'm actually getting errors from this. It looks like `go mod init` is still being called, and because this file already exists, it throws `go.mod already exists`

generators/newapp/new.go
```diff
@@ -86,12 +68,25 @@ func (a Generator) Run(root string, data makr.Data) error {
...
+			// This step needs to be in a separate generator, because goGet() exec.Command
+			// checks if the executable exists (so before running the generator).
+			gg := makr.New()
+			gg.Add(makr.NewCommand(makr.GoGet("github.com/golang/dep/cmd/dep")))
+			if err := gg.Run(root, data); err != nil {
+				return errors.WithStack(err)
+			}
+		}
+	} else {
+		if _, err := exec.LookPath("goimports"); err != nil {
```
stanislas-m: If I use dep, will goimports still be installed?

markbates: lol. fixed.

PR #1256, opened 2018-08-24
Adding alias to middleware imports

paganotoni:


robbyoconnor:
You might also need to update the code, no?

PR #1259, opened 2018-08-25
Fix middlewares on update

paganotoni:
This PR is part of the movement of the middlewares separate packages, this will make the fix/upgrade command update those imports.


buffalo/cmd/fix/middleware.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return err
+	}
+
+	newContents := string(read)
+	newContents = strings.Replace(newContents, "middleware.SetContentType", "contenttype.Set", -1)
+	newContents = strings.Replace(newContents, "middleware.AddContentType", "contenttype.Add", -1)
+	newContents = strings.Replace(newContents, "middleware.ParameterLogger", "paramlogger.ParameterLogger", -1)
+	newContents = strings.Replace(newContents, "middleware.PopTransaction", "popmw.Transaction", -1)
+
```
kteb: it's missing `newContents = string.Replace(newContents, "middleware.ForceSSL", "ssl.Middleware", -1)`

paganotoni: Oh, you are right, i only replaced the "root" middleware expressions but may need to also replace the ones using packages as well, i'll also add some tests.

paganotoni: Thanks for the review.

kteb:
Thanks it looks great, I tried it in an app and it went well this time ;).

PR #1261, opened 2018-08-26
Skip generating inflections.json if no pop

msabramo:


markbates:
Thanks, but other tools in the chain also use that file, so we definitely want to keep it around. Thanks though.

msabramo:
Ok no problem. Out of curiosity, since I’m new to Buffalo (but I really like what I see so far!), what things use that file?

markbates:
Anything that use inflection, so basically all of the generators. :)

Thanks.

-----------
Mark Bates

PR #1269, opened 2018-08-29
respect the publicPath from webpack

bittopaz:
Prefix this hard-coded "/assets" will break the public path logic from webpack.

If I specify the public path in webpack config with "https://cdn.mydomain.com/staticfiles/", this template helper will prefix "/assets" to "/assets/https://cdn.mydomain.com/staticfiles/", which break the CDN url for the asset.


render/template_helpers.go
```diff
@@ -26,9 +26,9 @@ func assetPathFor(file string) string {
 	filePath := assetMap[file]
 	assetsMutex.RUnlock()
 	if filePath == "" {
-		filePath = file
+		filePath = filepath.Join("/assets", file)
 	}
-	return filepath.ToSlash(filepath.Join("/assets", filePath))
+	return filepath.ToSlash(filepath)
```
stanislas-m: This line should be `return filepath.ToSlash(filePath)`. :)

PR #1276, opened 2018-09-01
Paramlogger on

paganotoni:
This PR attempts to make paramlogger ON for all environments now that it masks sensitive data.
see: https://github.com/gobuffalo/mw-paramlogger/blob/master/param_logger.go#L110

### DIFF
--- a/generators/newapp/templates/actions/app.go.tmpl
+++ b/generators/newapp/templates/actions/app.go.tmpl
@@ -58,17 +58,14 @@ func App() *buffalo.App {
       {{ end -}}
       SessionName: "_{{.opts.Name.File}}_session",
     })
+    
     // Automatically redirect to SSL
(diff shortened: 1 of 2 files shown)


generators/newapp/templates/actions/app.go.tmpl
```diff
@@ -58,18 +58,15 @@ func App() *buffalo.App {
       {{ end -}}
       SessionName: "_{{.opts.Name.File}}_session",
     })
+
     // Automatically redirect to SSL
-    app.Use(forceSSL())
+    app.Use(forceSSL(), paramlogger.ParameterLogger)
```
markbates: Can you split this into two lines and add a small comment explaining?

paganotoni:
@markbates comment added and split it.

PR #1283, opened 2018-09-02
[WIP] Adds fingerprinted images to the manifest

paganotoni:
@stanislas-m it seems i did find the way to have images being fingerprinted (yes our old story :D ), im opening this PR to ask for your help with other things we may need to change if we introduce this one, from the top of my head i have:

[] Update the index.html page to use the manifest (using the assetPath helper for images)


paganotoni:
@gobuffalo/core-managers this works good, my only fear with this one is we should provide some transition steps as existing html images will need to use `assetsPath()` helper, one thing i've been thinking is we could to make the upgrader change images from `href="assets/.*"` to assetPath.

I wonder if you guys have other ideas that could help make the transition smooth.

mclark4386:
Good point... that said I think the only projects this will affect are new (so not an issue) and those that pick to upgrade their webpack.config.js files in the upgrade process... so I think as long as we give good messaging that's the best we can do... there is no sure fire way to target all potential instances...

markbates:
@paganotoni what's the status of this PR? is it still WIP?

paganotoni:
It is ready, but i think we should provide a way to upgrade templates if developer chooses to update his webpack.yml file.

In that case we would replace all `/assets/[img]` srcs to use `assetsPath` helper instead of hardcoded URLs, otherwise image urls will break @markbates what do you think?

paganotoni:
Opening another PR with this changes.

PR #1293, opened 2018-09-07
Migrate Pop logger to its new signature

stanislas-m:


stanislas-m:
@markbates I don't think it's possible without a proper event system, since it requires the logger set as an option.

With an event system, we could have something like a `OnConfig` handler we can register on buffalo-pop side.

markbates:
That’s what I thought. But wanted to check.

PR #1294, opened 2018-09-07
Do not join absolute paths on build messages

agonzalezro:
`buffalo build` was showing the wrong output message when -o was provided with
an absolute path because it was appending it to b.Root when it shouldn't be doing
that.

Example of how it was:

```
$ buffalo build -o /tmp/test
You application was successfully built at /Users/alex/go/src/github.com/agonzalezro/buffalo/tmp/test
```

Example of how it is:

```
$ buffalo build -o /tmp/test
You application was successfully built at /tmp/test
```


agonzalezro:
I was trying to make the test more flexible using `path.Join` for the expected result instead writing it down, however I think it make it difficult to understand. What do you think about having a test for Windows systems and one for *nix system using build tags for this?

markbates:
you could use `filepath.Join` in the tests instead of `foo/bar`. I haven't looked at the errors, but I'm assuming its that based on the test code.

agonzalezro:
The "problem" if the tests are not separated for Windows and *nix is that an input binary path like this: `/foo/bar` in Windows should end up (I think, please correct me if I am wrong) in something like: `C:\\thisismycurrentpathexample\/foo/bar`. Instead, in *nix that would be an absolute path and it would end up at `/foo/bar`.

If we don't want to add logic to the tests the cleanest approach seems to have two different tests cases depending on the OS. I was asking the question because I saw build tags only for appengine and integration tests and maybe you don't want to follow that approach and I am overcomplicating something that should be simpler.

markbates:
Ah gotcha. We haven’t been using build tags for testing like that , but have been using runtime.GOOS for that check and just having two different tables in the same test. That way we don’t have to change the test logic but just the data we feed in/out.

--
Mark Bates

PR #1310, opened 2018-09-19
adds support for the default logger to emit events

markbates:


stanislas-m:
I'm not sure about this one, it looks like it will add a lot of overhead. Do you have a use-case for it? :)

markbates:
@stanislas-m I don't. I was wondering about this one myself, which is why I opened it as a separate PR. I'm going to close it. Two of us are questioning it's use, including me. :) It's easy enough for someone to whip up on their own.

Issue #1296, opened 2018-09-07
Mail renderer doesn't know routes


lukasschlueter:
## Steps to Reproduce the Problem

  Minimal setup:

  1. Create a new project
  1. Run `buffalo g mailer Test`
  1. Add SMTP credentials to `.env`
  1. Add handler that calls the new mailer, e.g. 
``` Go
app.GET("/send", func(context buffalo.Context) error {
	return mailers.SendTest()
})
```
  5. Specify mail receiver in `mailers/test.go`
  1. Edit `templates/mail/test.html` to include something like `Website root : "<%= rootPath() %>"`
  1. Open `localhost:3000/send`

## Expected Behavior

Buffalo sends a mail to the specified email address containing `Website root: "/"`.

## Actual Behavior

Buffalo fails to render the template with the following error: `test.html: line 3: rootPath: unknown identifier`

## Info

<details>

```
### Buffalo Version
v0.12.6

### App Information
Pwd=C:\Users\Lukas\go\src\gitlab.com\lukas-schlueter\test
Root=C:\Users\Lukas\go\src\gitlab.com\lukas-schlueter\test
GoPath=C:\Users\Lukas\go
Name=license_server
Bin=bin\license_server.exe
PackagePkg=gitlab.com/lukas-schlueter/test
ActionsPkg=gitlab.com/lukas-schlueter/test/actions
ModelsPkg=gitlab.com/lukas-schlueter/test/models
GriftsPkg=gitlab.com/lukas-schlueter/test/grifts
VCS=git
WithPop=true
WithSQLite=true
WithDep=false
WithWebpack=true
WithYarn=true
WithDocker=true
WithGrifts=true

### Go Version
go version go1.11 windows/amd64

### Go Env
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\Lukas\AppData\Local\go-build
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=C:\Users\Lukas\go
set GOPROXY=
set GORACE=
set GOROOT=D:\Programs\Go
set GOTMPDIR=
set GOTOOLDIR=D:\Programs\Go\pkg\tool\windows_amd64
set GCCGO=gccgo
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=C:\Users\Lukas\go\src\gitlab.com\lukas-schlueter\test\go.mod
[...]

PR #1324, opened 2018-09-26
[WIP] access routes from email

kteb:
First try of #1296 

I tried quickly the code and it seems working, but I love a review on this one. 

And advice where to add the the tests (maybe in `default_context_test.go` ? 

The changes in the mailer are breaking changes, so I think adaption in `buffalo fix` is needed (it will not stop working, but will not give this functionality to the people who used the old generator).


kteb:
Better implementation on #1338

PR #1338, opened 2018-09-30
adds a couple of new constructors for Message, including New(*buffalo.App). this fixes #1296

### LINKED ISSUE #1296: Mail renderer doesn't know routes

markbates:
@kteb I know you took a swing at this too. There were a things about your approach that didn't feel "right" to me. 

My approach adds two new attributes to `Message` : `Context Context` and `Data render.Data`. When calls to functions like `Message#AddBody` the `Message.Data` and the passed in `data` will be merged together.

I've also add a `New(c buffalo.Context) Message`. We should recommend people use this function when sending emails from an action. This will add the context's data  to the `Message.Data`. This brings along the route paths.

I'm up for discussion on this if people don't like it.


kteb:
It looks nice. Quick question about the mutex, if I'm right, it here in case we have a process sending multiple emails, and we don't want to mix the data field between them, right ? 

Shouldn't we move this package to a plugin ?

markbates:
I get paranoid around working with maps since they’re not thread safe, hence the mutex. I don’t see a lot of times that that code would be run concurrently, but if it is, it won’t break. :)

And yes, mail probably should be a plug-in, but one thing at a time. :)

Thanks.

-----------
Mark Bates

PR #1344, opened 2018-10-01
Feature: Docker Compose

rxx:
2 versions of docker-compose

* deps
contains only database container
* full
contains database container and app container

### DIFF
--- a/buffalo/cmd/new.go
+++ b/buffalo/cmd/new.go
@@ -31,18 +31,19 @@ var configError error
 func getAppWithConfig() newapp.Generator {
 	pwd, _ := os.Getwd()
 	app := newapp.Generator{
-		App:         meta.New(pwd),
-		AsAPI:       viper.GetBool("api"),
-		Force:       viper.GetBool("force"),
-		Verbose:     viper.GetBool("verbose"),
-		SkipPop:     viper.GetBool("skip-pop"),
-		SkipWebpack: viper.GetBool("skip-webpack"),
-		SkipYarn:    viper.GetBool("skip-yarn"),
-		DBType:      viper.GetString("db-type"),
-		CIProvider:  viper.GetString("ci-provider"),
-		AsWeb:       true,
-		Docker:      viper.GetString("docker"),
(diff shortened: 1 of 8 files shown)


rxx:
@markbates One problem. It does not work on `buffalo db` namespace. It does not run my preRun hook. Could you help?

markbates:
@rxx thanks for the PR, unfortunately, I don't think it's a) the correct pattern and b) belong in "core". Please let me explain, as well, as my suggestions on how to proceed, because I think the idea is great!

First the pattern. As you currently have it regardless of what command you run it'll spin up docker-compose. That seems problematic for a variety of reasons. It also seems quite heavy handed if you had no intention to start, nor need it run. I would suggest a command that would start allow users to start/stop docker compose in the background.

As for where it belongs, we're moving in a pattern that is trying to shed features, such as docker, pop, even `buffalo dev` will eventually be moved to a plugin. The reasons for this are laid out in a series of posts on https://blog.gobuffalo.io. This is also the reason you were having trouble with it running with `buffalo db`, that is now a plugin, so isn't being run with your pre-run.

So my suggestion would be to create a plugin, `buffalo-docker` and it put all this functionality in there. It has been on my todo list for a while now to move the docker generators out of "core" and into a `buffalo-docker` plugin.

The https://github.com/gobuffalo/buffalo-plugins plugin has a great generator to get your started, just look at it's help `-h` for more details. Almost all of the `gobuffalo/buffalo-*` plugins were generated using this plugin, so they're good examples of how to write them.

Here's the godoc for the generator framework we use https://godoc.org/github.com/gobuffalo/genny

And finally a link to the branch I have where I rewrote the docker generates using genny. https://github.com/gobuffalo/buffalo/tree/rewrite-a-few-generators/genny/docker feel free and copy that code right out of there. You'd be doing me a favor. :)

Create the plugin on your account, and when you think it's ready open an issue and ping me. 

I hope you understand, and thanks for the help!

rxx:
@markbates Thanks for pointing me the right way. I will on plugin system more deeply

PR #1345, opened 2018-10-01
Array already has time.RFC3339 const

rxx:
I found that raw date at timeFormats is exactly the `time.RFC3339`, so we can remove duplicate one


rxx:
@markbates I've updated

markbates:
Thanks!

PR #1351, opened 2018-10-04
Better handle errors display as JSON or XML

stanislas-m:
* Make sure using the XML mime type returns an XML-formatted error.
* Use the same struct for XML & JSON.
* Add a new attribute to have only the cause and not the full trace.


stanislas-m:
Fixes  #1353.

PR #1354, opened 2018-10-05
Extracting module name from go.mod

paganotoni:
If using go modules we should get the package name of the app from the go.mod file instead of getting it from the path.


paganotoni:
@markbates added some tests

PR #1381, opened 2018-10-18
removes the docker generator in favor of the buffalo-docker plugin

markbates:


generators/newapp/new.go
```diff
@@ -98,12 +99,61 @@ func (a Generator) Run(root string, data makr.Data) error {
...
+
 	a.setupVCS(g)
 
 	data["opts"] = a
 	return g.Run(root, data)
 }
 
+func (a Generator) genny() (makr.Runnable, error) {
+	app := a.App
+	run := genny.WetRunner(context.Background())
```
gangachris: @markbates I noticed the current development branch generates Dockerfile and .dockerignore outside the project directory. 

For example, this generates the Dockerfiles in `../petstore`
```
buffalo new petstore --api -f
```

I tracked this down to simply adding a line after this `run.Root = app.Root`, but I'm not entirely sure this is idiomatic. I'm thinking WetRunner should take in a Path as a variable. 

What do you think?

markbates:
We should set the Run.Root, but my guess is the docker generator isn’t doing to right and should be fixed as well.

--
Mark Bates
CTO/Founder

gangachris:
Should I add the `run.Root` where suggested, or you'll fix it in the Generator?

markbates:
@gangachris the `run.Root` fix is already in the `development` branch https://github.com/gobuffalo/buffalo/blob/development/generators/newapp/new.go#L121

So I was suggesting opening a PR to buffalo-docker to make sure it sets it's own root, if given.

Issue #1153, opened 2018-06-30
Proposal: Add a standard `config` directory, format, and naming structure


markbates:
As Buffalo works to make tooling and plugins easier to write, there is going to become a need for these tools to be able to configure themselves, and to configure them for use with Buffalo.

It is also helpful to app developers to have a standard place to put their configs for their apps as well.

To this end I propose a new `config/` folder in applications. This folder will automatically be bundled with `buffalo build` so tools will still have access to them at right time. app developers can also access them as well using `packr.NewBox("<path to config/>").

I believe all need a common file format and naming system. I propose TOML for the file format and `config/<tool-name>.toml` for the file name. 

**"External" Tools:**
Example `config/vendor/buffalo-heroku.toml`.
Example `config/vendor/buffalo-azure.toml`.

**"Buffalo/Internal" Tools:**
Example `config/buffalo/dev.toml`.
Example `config/buffalo/build.toml`.

**Reserved names in `config/` are**:
`app.toml`
`buffalo.toml`

gedw99:
Like..

So the env approach works in tandem with this ? Or they are for different aspects ?

markbates:
Yes, the `.env` is for configuring environment variables. this folder is meant mostly for tool developers to have a common place to configurations for their tools. think, for example, `buffalo dev` or `buffalo test`, etc... 

It would also, but it being there, would be a place that app developers could leverage to put any configuration files they might want.

gedw99:
Got it. Thanks Mark

PR #1382, opened 2018-10-18
encodes meta.App to config/buffalo-app.toml on buffalo new

markbates:
```bash
$ buffalo new coke -f --api --with-dep
```

```toml
# ./config/buffalo-app.toml
name = "coke"
bin = "bin/coke"
vcs = "git"
with_pop = true
with_sqlite = false
with_dep = true
with_webpack = false
with_yarn = true
with_docker = true
with_grifts = false
as_web = false
as_api = true
```

Changes made to `config/buffalo-app.toml` will be picked up by tooling that uses `github.com/gobuffalo/meta#App`.

If this file does not exist, then it will fall back to the current behavior of trying to auto detect.

This PR will also allow for us to change generators to spit out API templates instead of HTML, for example, if `AsAPI` is set.


markbates:
Related to https://github.com/gobuffalo/buffalo/issues/1153

PR #1383, opened 2018-10-18
Fix a race condition in DefaultContext.data

duckbrain:
There were `RLock` calls around all the accesses of the `DefaultContext.data` map, but no `Lock` around the write in the `Set` method. This adds that `Lock`. 

Updating to v0.13.0 caused my application to start panicing with `fatal error: concurrent map iteration and map write`. While writing up the issue, I found the problem.

I'd recommend considering to make this a patch update for 0.13.0.


markbates:
Thank you. This is already fixed in the development branch. 

https://github.com/gobuffalo/buffalo/blob/development/default_context.go

markbates:
And yes. I’m prepping a release for tomorrow.

PR #1388, opened 2018-10-19
add documentation for contributing with go modules

gangachris:
Since support for Go Modules was introduced, I've added documentation on how to start contributing if you are already using go modules.


gangachris:
@markbates I'm curious what you think about this since the current linter gometalinter does not support Go Modules yet.

Issue #1353, opened 2018-10-05
Better handle errors display as JSON or XML


stanislas-m:
The current default error handler doesn't handle the case when the mime type is application/xml. We should display errors in XML, just like we do with JSON.

Stack trace for errors are useful, but some folks on the Slack expects to be able to have the root cause directly. We should split the current `error` attribute so we'll have `error` (the root error / cause) and `trace` in a separate attribute.

cippaciong:
Hello, in the context of a project created with the `--api` flag, with `app.Use(contenttype.Set("application/json"))` and `DefaultContentType: "application/json"` set in rendered options, after switching from development mode to production, errors are not returned in JSON but as HTML pages. I think it would be nice to heave "default" JSON/XML errors in production just as we have error pages. What do you think about it?

stanislas-m:
Sure, why not. Do you want to submit a PR for that, @cippaciong?

cippaciong:
Yeah, I can give it a try in the next days.

PR #1395, opened 2018-10-21
Add default error response for json/xml in production env

pzeinlinger:
Addresses the second comment in #1353 by cippaciong.
If generating a new buffalo project with the `--api` flag and `app.Use(contenttype.Set("application/json"))` is used, errors were not returned in JSON but as HTML pages in production.
This commit fixes this problem.


errors.go
```diff
@@ -140,6 +140,13 @@ type ErrorResponse struct {
 	Code    int      `json:"code" xml:"code,attr"`
 }
 
+// ErrorResponseProduction is a used to display errors as JSON or XML in production
+type ErrorResponseProduction struct {
```
markbates: I don’t see a need for a second almost identical structure. If trace is unwanted in production either it shouldn’t be set in production or under before rendering

markbates:
@stanislas-m I’ll let you own this since you wrote the original. I’ve provided some feedback already.

gangachris:
@pzeinlinger any idea why travis build fails here 🤔 
```
......
github.com/dgrijalva/jwt-go (download)
make: *** [ci-deps] Error 1
```

pzeinlinger:
@stanislas-m I'm sorry, but I disagree with your last message. A JSON-APIs should provide verbose errors to its users. You can find numerous articles around the web that all mention being verbose being a good thing in the context of JSON-APIs (e.g. https://medium.com/studioarmix/learn-restful-api-design-ideals-c5ec915a430f)

I can see why you wouldn't want to do the same thing for HTML(text) requests that are usually targeted for non-developers/customers.

stanislas-m:
While I agree verbose messages can be useful, I think there's a difference between technical errors and functional errors. Technical ones can leak precious info like db host, and need to be neutralized. Functional errors are not the same errors, and can just be handled as normal structs. Counting on Go errors to return functional errors is abusing the system, in my opinion.

errors.go
```diff
@@ -149,8 +149,27 @@ func defaultErrorHandler(status int, origErr error, c Context) error {
...
-		responseBody := productionErrorResponseFor(status)
-		c.Response().Write(responseBody)
+		switch strings.ToLower(ct) {
+		case "application/json", "text/json", "json":
+			err := json.NewEncoder(c.Response()).Encode(&ErrorResponse{
+				Error: errors.Cause(origErr).Error(),
+				Code:  status,
+			})
+			if err != nil {
+				return errors.WithStack(err)
```
robsliwi: `WithStack` shouldn't be part of the response. Imagine bein a consumer of an API that is used in production and it tells you a lot of unneeded Trace. Let it aside that it could be even dangerous to leak such information to exposed world.

pzeinlinger: Is it? Wouldn't it just call the default error handler another time, just overwriting the initial error and not leaking the trace since it's in production? Do I miss anything here?
But I guess, since we wouldn't use the added trace, we could also just omit it and just return the new err.

PR #1397, opened 2018-10-21
dont add a go.mo file when go modules are turned off

gangachris:
go.mod file was being generated even when Go modules are turned off.


markbates:
Thanks!

PR #1403, opened 2018-10-23
Negotiate Response Content-Type from request's Accept header.

### LINKED ISSUE #1148: Correct Content-Type/Accept header usage
I don't understand the purpose of SetContentType for API's. It overrides the *request's* content type. [#1093](https://github.com/gobuffalo/buffalo/issues/1093#issuecomment-395971816) Makes me think maybe the `Accept` header was intended.

Anyway, this was causing problems with File uploads because the request's content type was always `application/json` (for PUT requests, but not POST requests?), not `multipart/form-data`. Changing it to AddContentType makes it work fine.

I have a patch ready to go to replace it with AddContentType (it's 3 characters), but I wanted some clarification on the purpose before sending a pull request. AddContentType makes more sense to me because it still respects the client's type and doesn't cause conflicts with File uploads and such.

## Steps to Reproduce the Problem

  Please describe in painful detail what you did (so others can play along with you) to get to this point. This includes things like the exact command(s) you used, or the `curl` command you used, that sort of thing.

  1. Generate a new project with the `--api` flag
  1. Add a PUT route with a file upload
  1. Start the server and upload a file

```bash
buffalo new test --api --skip-pop; cd test
buffalo g action upload UploadTest --method PUT --skip-template
vim actions/upload.go
buffalo dev
```

New contents for actions/upload.go

```go
package actions

import "github.com/gobuffalo/buffalo"

// UploadUploadTest default implementation.
func UploadUploadTest(c buffalo.Context) error {
	c.Logger().Debug(c.Request().Header.Get("Content-Type"))
	if _, err := c.File("file"); err != nil {
		return err
	}
	return c.Render(200, r.HTML("upload/upload_test.html"))
}
```

## Expected Behavior

I expect that it fails to display upload/upload_test.html after succeeding to parse the uploaded file

## Actual Behavior

It logs `application/json` to the debug log and errors out with `request Content-Type isn't multipart/form-data`

If I go into `actions/app.go` and change the middleware out, it starts working.

## Info

<details>

```
### Buffalo Version
v0.12.3
### App Information
Pwd=/home/jonathan/src/git.ec2software.com/jonathan/test
Root=/home/jonathan/src/git.ec2software.com/jonathan/test
GoPath=/home/jonathan
Name=test
Bin=bin/test
PackagePkg=git.ec2software.com/jonathan/test
ActionsPkg=git.ec2software.com/jonathan/test/actions
ModelsPkg=git.ec2software.com/jonathan/test/models
GriftsPkg=git.ec2software.com/jonathan/test/grifts
VCS=git
WithPop=false
WithSQLite=false
WithDep=false
WithWebpack=false
WithYarn=false
WithDocker=true
WithGrifts=true
### Go Version
go version go1.10.3 linux/amd64
### Go Env
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/jonathan/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/jonathan"
GORACE=""
GOROOT="/usr/lib/go"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build564953439=/tmp/go-build -gno-record-gcc-switches"
### Node Version
v9.11.2
### NPM Version
5.6.0
### Yarn Version
1.7.0
### PostgreSQL Version
PostgreSQL Not Found
### MySQL Version
mysql  Ver 15.1 Distrib 10.1.34-MariaDB, for Linux (x86_64) using readline 5.1
### SQLite Version
3.24.0 2018-06-04 19:24:41 c7ee0833225bfd8c5ec2f9bf62b97c4e04d03bd9566366d5221ac8fb199a87ca
### Dep Version
could not find a Gopkg.toml file
### Dep Status
could not find a Gopkg.toml file
```

</details>

duckbrain:
Resolves #1148

This changes the default Context to negotiate the response `Content-Type` using the request's `Accept` header exclusively, instead of including the request's `Content-Type` in the negotiation. Failed negotiation will default to using `render.Options#DefaultContentType`. (This is important because Internet Explorer does not pass an Accept header that negotiates HTML.)

### This is not backwards compatible

HTTP clients using `Content-Type` instead of `Accept` in their request headers will not work the same way. Code that overrides these values (such as those using https://github.com/gobuffalo/mw-contenttype) will no longer have the desired effect. They should be removed in favor of the `DefaultContentType` option or a middleware that sets the `Accept` header.

~~https://github.com/gobuffalo/httptest/pull/2 is needed for tests to pass~~ Merged!

### DIFF
--- a/binding/binding.go
+++ b/binding/binding.go
@@ -82,7 +82,7 @@ func RegisteredTypes() []string {
 	defer lock.Unlock()
 
 	types := []string{}
-	for name, _ := range binders {
+	for name := range binders {
 		types = append(types, name)
 	}
 	return types
--- a/binding/binding.go
+++ b/binding/binding.go
@@ -75,6 +75,19 @@ func Register(contentType string, fn Binder) {
 	binders[strings.ToLower(contentType)] = fn
 }
 
+// RegisteredTypes gets a list of the Content-Types that have been registered
+// with Binders and can be used in negotiating a response Content-Type.
+func RegisteredTypes() []string {
+	lock.Lock()
+	defer lock.Unlock()
+
+	types := []string{}
+	for name, _ := range binders {
+		types = append(types, name)
+	}
+	return types
+}
+
 // Exec will bind the interface to the request.Body. The type of binding
 // is dependent on the "Content-Type" for the request. If the type
 // is "application/json" it will use "json.NewDecoder". If the type
--- a/context.go
+++ b/context.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/gobuffalo/buffalo/binding"
 	"github.com/gobuffalo/buffalo/render"
-	"github.com/gobuffalo/x/httpx"
+	"github.com/golang/gddo/httputil"
 	"github.com/gorilla/mux"
 )
 
@@ -62,7 +62,7 @@ func (a *App) newContext(info RouteInfo, res http.ResponseWriter, req *http.Requ
 
 	session := a.getSession(req, res)
 
-	ct := httpx.ContentType(req)
+	ct := httputil.NegotiateContentType(req, binding.RegisteredTypes(), "")
 	contextData := map[string]interface{}{
 		"app":           a,
 		"env":           a.Env,


duckbrain:
Does this need to add something for `buffalo fix` to fix https://github.com/gobuffalo/mw-contenttype middleware links to a new route that `Add/Set`'s the `Accept` header?

PR #1409, opened 2018-10-25
Fix outdated imports of gobuffalo/buffalo/meta

cippaciong:
There were some imports of `github.com/gobuffalo/buffalo/meta` that were not updated to use the new `github.com/gobuffalo/meta` causing this message to come up when running `buffalo`:
```
[DEPRECATED] github.com/gobuffalo/buffalo/meta has been deprecated.
    Use github.com/gobuffalo/meta instead.
```


markbates:
Thanks. This is fixed in #1396.

Issue #1406, opened 2018-10-24
context not passed when `partial` used within `for` loop


sio4:
When I tried to migrate my old app into the current version of buffalo, I found the following problem.

I guess context not passed to `partial` when it called within `for` loop.

## Description

When I tried to use `partial()` within `for` loop, it cannot find the parent's context variables.
It did just work fine with 0.11.0 (as I remember) but now it does not work.

### Steps to Reproduce the Problem

parent template:

```
    <%= for (user) in users { %>
      <%= partial("users/row") %>
    <% } %>
```

a partial template named `_row.html`

```
<%= user %>
```

Error message:

```
users/index.html: line 25: users/_row.html: line 2: user: unknown identifier
```

If I use `<% let user = "something" %>`, it passed well to partial but if I use `let` inside of the loop, it is also not working.

### Expected Behavior

Render content with the parent's value.

### Actual Behavior

make error: unknown identifier

### Info

<details>

```
### Buffalo Version
development

### App Information
Pwd=/home/sio4/gowork/src/github.com/hyeoncheon/skel
Root=/home/sio4/gowork/src/github.com/hyeoncheon/skel
GoPath=/home/sio4/gowork
Name=skel
Bin=bin/skel
PackagePkg=github.com/hyeoncheon/skel
ActionsPkg=github.com/hyeoncheon/skel/actions
ModelsPkg=github.com/hyeoncheon/skel/models
GriftsPkg=github.com/hyeoncheon/skel/grifts
VCS=git
WithPop=true
WithSQLite=false
WithDep=false
WithWebpack=true
WithYarn=true
WithDocker=true
WithGrifts=true
WithModules=false

### Go Version
go version go1.11.1 linux/amd64

### Go Env
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/sio4/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/sio4/gowork"
GOPROXY=""
GORACE=""
GOROOT="/opt/google/go"
GOTMPDIR=""
GOTOOLDIR="/opt/google/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
[...]

PR #1411, opened 2018-10-26
integrated with plush's own partial (and added testcases)

sio4:
Hi, @markbates and all,

Added a test case for issue #1406. I think the solution to this issue should be in `plush` and I tested with patch below:

```diff
--- a/compiler.go
+++ b/compiler.go
@@ -504,6 +504,21 @@ func (c *compiler) evalCallExpression(node *ast.CallExpression) (interface{}, er
                                return nil, errors.WithStack(err)
                        }
 
+                       // 3. if the function is `partial` and user provides second argument,
+                       //    merge it with current context's data which has loop var.
+                       if node.Function.String() == "partial" && pos == 1 {
+                               fmt.Printf("XXX - %T, %v\n", v, v)
+                               data := map[string]interface{}{}
+                               for k, vv := range c.ctx.data {
+                                       data[k] = vv
+                               }
+                               for k, vv := range v.(map[string]interface{}) {
+                                       data[k] = vv
+                               }
+                               v = data
+                               fmt.Printf("XXX - %T, %v\n", v, v)
+                       }
+
                        var ar reflect.Value
                        expectedT := rt.In(pos)
                        if v != nil {
@@ -520,6 +535,12 @@ func (c *compiler) evalCallExpression(node *ast.CallExpression) (interface{}, er
                        args = append(args, ar)
                }
 
+               // 3.1. if the function is `partial` but there is no second argument,
+               //      just use current context's data as second argument.
+               if node.Function.String() == "partial" && len(node.Arguments) == 1 {
+                       args = append(args, reflect.ValueOf(c.ctx.data))
+               }
+
                hc := func(arg reflect.Type) {
                        if arg.ConvertibleTo(reflect.TypeOf(HelperContext{})) {
                                hargs := HelperContext{
@@ -619,6 +640,7 @@ func (c *compiler) evalCallExpression(node *ast.CallExpression) (interface{}, er
                }
        }
 
+       // 4. finally it calls `partial()` but it runs on another context.
        res := rv.Call(args)
        if len(res) > 0 {
                if e, ok := res[len(res)-1].Interface().(error); ok {
@@ -634,6 +656,8 @@ func (c *compiler) evalForExpression(node *ast.ForExpression) (interface{}, erro
        defer func() {
                c.ctx = octx
        }()
+       // 1. create new context for `for` block here.
+       //    anyway this context does not have values form original.
        c.ctx = octx.New()
        iter, err := c.evalExpression(node.Iterable)
        if err != nil {
@@ -661,8 +685,9 @@ func (c *compiler) evalForExpression(node *ast.ForExpression) (interface{}, erro
        case reflect.Slice, reflect.Array:
                for i := 0; i < riter.Len(); i++ {
                        v := riter.Index(i)
-                       c.ctx.Set(node.KeyName, i)
+                       c.ctx.Set(node.KeyName, i) // misc: is it required?
                        c.ctx.Set(node.ValueName, v.Interface())
+                       // 2. it calls evalBlockStatement after set new values
                        res, err := c.evalBlockStatement(node.Block)
                        if err != nil {
                                return nil, errors.WithStack(err)
```

### DIFF
--- a/render/js.go
+++ b/render/js.go
@@ -1,11 +1,8 @@
 package render
 
 import (
-	"html/template"
-	"strings"
-
 	"github.com/gobuffalo/plush"
-	"github.com/pkg/errors"
+	"github.com/markbates/oncer"
 )
 
 // JavaScript renders the named files using the 'application/javascript'
@@ -41,28 +38,8 @@ func (e *Engine) JavaScript(names ...string) Renderer {
 }
 
 // JSTemplateEngine renders files with a `.js` extension through Plush.
-// It also implements a new `partial` helper that will run non-JS partials
-// through `JSEscapeString` before injecting.
+// Deprecated: use github.com/gobuffalo/plush.BuffaloRenderer instead.
 func JSTemplateEngine(input string, data map[string]interface{}, helpers map[string]interface{}) (string, error) {
-	var pf partFunc
-	var ok bool
-	if pf, ok = helpers["partial"].(func(string, Data) (template.HTML, error)); !ok {
-		return "", errors.New("could not find a partial function")
-	}
-
-	helpers["partial"] = func(name string, dd Data) (template.HTML, error) {
-		if strings.Contains(name, ".js") {
-			return pf(name, dd)
-		}
-		h, err := pf(name, dd)
-		if err != nil {
-			return "", errors.WithStack(err)
-		}
-		he := template.JSEscapeString(string(h))
-		return template.HTML(he), nil
-	}
-
+	oncer.Deprecate(0, "render.JSTemplateEngine", "Use github.com/gobuffalo/plush.BuffaloRenderer instead.")
 	return plush.BuffaloRenderer(input, data, helpers)
 }
-
-type partFunc func(string, Data) (template.HTML, error)
--- a/render/js_test.go
+++ b/render/js_test.go
@@ -1,4 +1,4 @@
-package render_test
+package render
 
 import (
 	"bytes"
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/gobuffalo/buffalo/render"
 	"github.com/gobuffalo/packr"
 	"github.com/stretchr/testify/require"
 )
@@ -30,7 +29,7 @@ func Test_JavaScript(t *testing.T) {
 	t.Run("without a layout", func(st *testing.T) {
 		r := require.New(st)
 
-		j := render.New(render.Options{
+		j := New(Options{
 			TemplatesBox: packr.NewBox(tmpDir),
 		}).JavaScript
 
@@ -51,7 +50,7 @@ func Test_JavaScript(t *testing.T) {
 		_, err = layout.Write([]byte("<body><%= yield %></body>"))
 		r.NoError(err)
 
-		re := render.New(render.Options{
+		re := New(Options{
 			JavaScriptLayout: filepath.Base(layout.Name()),
 			TemplatesBox:     packr.NewBox(tmpDir),
 		})
@@ -94,7 +93,7 @@ func Test_JavaScript_JS_Partial(t *testing.T) {
 	r.NoError(err)
 	defer os.RemoveAll(dir)
 
-	re := render.New(render.Options{
+	re := New(Options{
 		TemplatesBox: packr.NewBox(dir),
 	})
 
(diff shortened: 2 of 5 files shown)


sio4:
Travis and AppVeyor failed on these test cases. To solve issue #1406, it should be passed.

markbates:
@sio4 this is great! 

Thinking about what you've dug up, I agree that Plush should have a `partial` helper. It should have the signature as the one in Buffalo, but the different is the string it takes isn't a file name, but the HTML to be rendered. Then in Buffalo we can wrap that in our own `partial` function that gets the file, and sends it all onto the plush one. 

Does that sound right to you? If so, I would love it if you could own this. :)

sio4:
Hi @markbates 
I am sorry but I cannot fully understand your comment. (since my English is very poor)

Anyway, This PR is for testing the `partial` is working or not within `for` loop. If you tried this tests, or just simply checking the Travis-CI's output with the test case codes in this commit, you can find the issue still exists although you did patch. (The point of this PR and my comments is not about a position of `partial()` but the problem.)

For the implementation of partial, as I already wrote as comments on issue #1406,

> https://github.com/gobuffalo/buffalo/issues/1406#issuecomment-433334555

I understood why `partial()` was implemented on Buffalo side since it needs to handle file i/o and understands directory hierarchy of the application, and `plush` just handle given `string` without files and directory paths. So I fully agree with the current "call-back style" structure.

But since it still not working, we need to fix it and I think the patch for `plush` on above comment is one of the possible approaches.

markbates:
No problem. :) I was agreeing 100% with you. 👍 

* fix plush
* add `partial(htmlBody string, options, ctx)` helper to plush
* change buffalo's `partial` to call plush's `partial` function with the template file read from disk as a `string`

does that help?

sio4:
Yes, I think that will help.
Additionally, to keep compatibility and to keep it simple, I think the context should be held by the compiler as current, and pass the function pointer to application's partial feeder so it can return to compiler's context.

Anyway, currently, that feature is broken status so we need to fix it first.

sio4:
@markbates Something like this?

In plush side, add this and register it as "partial" helper:

```golang
func partialHelper(name string, data map[string]interface{}, help HelperContext) (template.HTML, error) {
	for k, v := range data {
		help.Context.data[k] = v
	}

	f := reflect.ValueOf(help.Context.data["partialFeeder"])
	args := []reflect.Value{reflect.ValueOf(name)}
	res := f.Call(args)

	body, err := Render(res[0].Interface().(string), help.Context)

	if data["layout"] != nil {
		return partialHelper(
			data["layout"].(string),
			map[string]interface{}{"yield": template.HTML(body)},
			help,
		)
	}

	return template.HTML(body), err
}
```
(before call it from `for` loop, all context data should be copied from parent.)

and in application side, implement something like this and register as "partialFeeder" helper:

```golang
func (s templateRenderer) partialFeeder(name string) (string, error) {
       d, f := filepath.Split(name)
       name = filepath.Join(d, "_"+f)

       if filepath.Ext(name) == "" {
               name += ".html"
       }

       source, err := s.TemplatesBox.MustBytes(name)
       if err != nil {
               return "", err
       }

       return string(source), nil
}
```

I didn't see in details but it passed most test cases of buffalo's `render/` but some javascript related test cases are failed.

sio4:
Hi @markbates, how do you think about my approach? If you agree, I will start to code it.

Please remember, **the `partial` within `for` is not working currently!** It needs hotfix before we get the direction for the future!

markbates:
Hi @sio4 sorry I haven't responded earlier, but I've been busy, and I don't think this is as wide spread and urgent as you think, as no one else is complaining off it, however, I agree it needs fixing, and I would like to get it into a `v0.13.x` release.

A few questions:

* is this a breaking change? I don't think it is, but I want to confirm
* can we change the `partialFeeder` stuff to *not* use reflection? you can use something like `pf, ok := data['feeder'].(plush.Feeder)` not proper code, but hopefully you understand.

If this isn't a breaking change, then please, start ASAP. :)

sio4:
Hi @markbates,
For your first questions, I also think that it should not affect to existing user's applications. it is not breaking change in point of user's applications. but since it is related to both Buffalo and Plush, and I think calling flow should be changed as the idea I shared, it needs to versioning as you always did and changed seamlessly on Buffalo side and Plush side.
For the second one, OK, I will try it to make it simpler as you said.

Let's start!

sio4:
I pushed two PRs on buffalo and plush. PR on plush is just adding a new helper and this PR uses that change. (so currently related test cases are failed on CIs, but I tested it locally.)

As @markbates asked, these changes do not affect other applications using plush but buffalo should be updated with this PR. I also removed codes using reflection and changed with as your recommendation.

Please check this.

markbates:
@sio4 can you rebase this against `master`? I want to cut a `v0.13.x` release with this, but `development` is not ready to be released. Thanks.

sio4:
Sure, I will do that by today.
Since now I am on the road, just after an hour or two.

-- 
...and in the end, the love you take is equal to the love make

2018년 10월 31일 (수) 22:33, Mark Bates <notifications@github.com>님이 작성:

markbates:
no rush. :) i know its the end of day for you.

-----------
Mark Bates

sio4:
Hi @markbates, I pushed new commits with the same changes against `master`, and just after the push, I changed the base branch. Maybe that reason, CI builds are failed. Can you trigger them manually?

markbates:
I briefly closed/re-opened the PR to see if that will trigger CI

PR #1434, opened 2018-11-02
Reset content-type to html for detailed error page

hdm:
This PR resolves a bug with the default development error handler where it will reflect the Request Content-Type. If a form handler triggers an error in development mode, the response Content-Type will be set to "application/x-www-form-urlencoded", which is wrong and causes the browser to display an error. The code responsible for this is `defaultErrorHandler()` in errorrs.go:

```
func defaultErrorHandler(status int, origErr error, c Context) error {
	env := c.Value("env")
	ct := defaults.String(httpx.ContentType(c.Request()), defErrorCT)
```

Prior to this fix, calling `return c.Error(404, err)` within a POST/PUT handler resulted in the following response:
```
< HTTP/1.1 404 Not Found
< Content-Type: application/x-www-form-urlencoded
< Date: Thu, 01 Nov 2018 21:46:15 GMT
< Transfer-Encoding: chunked
<
<html>
<head>
  <title>404 - ERROR!</title>
```

After this fix, the Content-Type is correctly set to: `text/html; charset=utf-8` and the error page renders correctly within the browser.


hdm:
This has been rebased against master as PR #1436

PR #1448, opened 2018-11-09
[WIP] Replace hardcoded webpack with std scripts

### LINKED ISSUE #1116: Proposal: Support for yarn builds

stanislas-m:
This PR replaces the hardcoded webpack commands (dev & build) with standard package.json scripts which can be called with either `npm` or `yarn`.

Fixes #1116


stanislas-m:
I'll open a better PR for that later.

PR #1450, opened 2018-11-10
prints a help message when buffalo-pop isn't installed but `buffalo db` is called

markbates:


mail/smtp_sender.go
```diff
@@ -5,7 +5,7 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
-	gomail "gopkg.in/mail.v2"
+	gomail "gopkg.in/gomail.v2"
```
stanislas-m:
```suggestion
	gomail "github.com/go-mail/mail/v2"
```

stanislas-m: gomail.v2 is the original repo, mail.v2 is the fork. mail.v2 points to `github.com/go-mail/mail`, and the maintainer suggests to use this version if we're using versioning tools.

PR #1451, opened 2018-11-11
Fix go-mail/mail path

stanislas-m:
go-mail/mail got updated with a go.mod file with a v2 path. This breaks the
new app generator when using `GO111MODULE=on`


markbates:
I fixed this is https://github.com/gobuffalo/buffalo/pull/1450 I want to release 0.13.5 tomorrow with the fix, as well as updated deps, and the pop message. If you can review for tomorrow.

markbates:
"fixed" in https://github.com/gobuffalo/buffalo/pull/1450

Issue #1440, opened 2018-05-26
model name singular to plural is not correct


netqyq:
potato's plural should be potatoes. 
```
# buffalo g resource potato title:string description:text --skip-templates
      create  actions/potatos.go
      create  actions/potatos_test.go
      create  locales/potatos.en-us.yaml
         run  buffalo db g model potato title:string description:text
v4.5.9

      create  models/potato.go
      create  models/potato_test.go
         run  goimports -w actions/actions_test.go actions/app.go actions/home.go actions/home_test.go actions/potatos.go actions/potatos_test.go actions/render.go actions/users.go grifts/db.go grifts/init.go main.go models/models.go models/models_test.go models/potato.go models/potato_test.go models/user.go models/user_test.go
      create  migrations/20180526135834_create_potatos.up.fizz
      create  migrations/20180526135834_create_potatos.down.fizz
         run  goimports -w actions/actions_test.go actions/app.go actions/home.go actions/home_test.go actions/potatos.go actions/potatos_test.go actions/render.go actions/users.go grifts/db.go grifts/init.go main.go models/models.go models/models_test.go models/potato.go models/potato_test.go models/user.go models/user_test.go

```

markbates:
This appears to be fixed now.

PR #1455, opened 2018-11-13
Improved generator using flect to handle plurals properly

sio4:
Improved generator to handle plurals properly, using `flect` instead of `inflect`.
It will solve issue #1440.

Please check the test case I fixed. I think it will work smoothly but I cannot sure since this is the first time I look into `flect` and this parts of codes.

Thanks.


sio4:
I fixed according to 'failed' `filetest` for nested resource and add normal test cases for it.
It can handle the input `admin/plane` to `admin_planes` and `admin/planes` and also handle the input `Admin/superFast/plane` to `admin/super_fast/planes` and `admin_super_fast_plane`. Is it desired result?

Please check the new test case and codes.

sio4:
Oh, I should not have touched it!
Its route is not `/admin/plane` :-(

sio4:
Hi @markbates,

While `ParamID() of inflect forces `Singular()` inside of the function, flect version of `ParamID()` just does `Underscore()` and add suffix (if not exists).

Currently, this PR does not work in point of adding proper routes and template autoloading but anyway Travis CI failure is on `ParamID()`. Can I fix `ParamID()` of flect? I think the way of inflect's is more reasonable.

sio4:
Hmm, Almost done but I cannot figure out how can I found templates automatically.

for routes, I just tried with `res.FilesPath` instead of `res.Name.URL()` since the underlying mechanism to generate them are same (but for flect, Ident removes '/' so I preserved it on `FilesPath` before).

But when I fall into the r.Auto(...), I have no idea to get paths of the templates since there is only `model`(`name` and `pname`) reflected from `data`. Any hints?

sio4:
Sorry for annoying comments during explorering. :-)

I fount function `File()` for `flect.name` lately and it solve several problems. But still, `URL()` of flect is not fit for nested resource something like `admin/planes`. So I think the following patch for flect is required.

```patch
--- a/name/url.go
+++ b/name/url.go
@@ -1,5 +1,5 @@
 package name
 
 func (n Ident) URL() Ident {
-       return Ident{n.Pluralize().Underscore()}
+       return Ident{n.File().Pluralize()}
 }
```

How do you think about this? Can I PR it before further progress?

sio4:
For template detection, is there any good approach with the current structure?
Currently, I have two Ideas

1) Pass context value named "template" before call `r.Auto()` and override with it when making template path.
2) Just preserve the current structure and handle it in template side: do not use `Auto()` but use `HTML()` for nested resources.

Any Ideas?

sio4:
Hi @markbates,

> The results of flect should match the results of inflect, so if you have to make changes there, go for it!
> 
> Please don’t wait for permission to open a PR, I would rather make suggestion on a PR than for you to wait for me to answer a question. 

OK, I opened PR for `flect`. Please check the PR. Since template feature uses the function directly, it is recommended that the PR need to be merged before this PR works properly.

> The answer to question 2 is leave. It as is. I investigated “true” nested resources ages ago, but the Go package structure makes it difficult as well.
> 
> Another this is the generator will be the next to be converted to genny, once I get the PR with the other generators merged. Don’t spend too much time improving this generator. Any change you make now will need to be ported to the genny implementation.

OMG! yes, I knew about your recent works on another branch! but I forget that and I just ran into this issue when I show the issue was labeled as `help wanted` :-) Anyway, I added one more commit to this PR and I think this is not a bad approach to handle this issue (and to handle some miss structured template directories or when we want to reuse existing template for other resources, of course, this is a rare case but I think it can happen for simple/similar resources)

Many thanks!

PR #1492, opened 2018-12-06
buffalo/cmd/build: forward flags to original main

rjeczalik:
Currently it is not possible to pass command line flags to a binary generated by `buffalo build`. User can face this problem when trying to parse flags in the main function, passing any extra flags on the command line breaks buffalo in different ways:

```
$ ./my-buffalo-app -my-flag value
ERRO[0001] Error: unknown command "value" for "buffalo"
 
2018/12/06 14:57:43 exit status 255
```

```
$ ./my-buffalo-app -my-flag=value
...

ERRO[0001] Error: unknown shorthand flag: 'm' in -my-flag=value
 
2018/12/06 14:56:34 exit status 255
```

```
$ ./my-buffalo-app -- -my-flag=value
(displays usage and exists with 0)
```

I propose making it possible to pass custom arguments to the user's main by respecting the `--` delimiter. In particular it would change the behaviour of 3rd example from the above - by executing `/my-buffalo-app -- -my-flag=value` buffalo is going to pass anything after the `--` delimiter to the user's main, that's `-my-flag=value`.

This PR introduces this change, ~~however it would be good to teach `buffalo dev` to also understand the `--` delimiter, so the following:~~

```
$ buffalo dev -- -my-flag=value
```

~~Works the same as:~~

```
$ buffalo build
$ ./bin/my-buffalo-app -- -my-flag=value
```

WDYT?


rjeczalik:
For the completeness I've added support for custom flags to the `buffalo dev` command, so the following:

```
$ buffalo dev -- -my-flag=value
```

Behaves exactly as:

```
$ buffalo build
$ ./bin/my-buffalo-app -- -my-flag=value
```

markbates:
Thanks!

PR #1493, opened 2018-12-07
adds packr2 support to v0.13.x

markbates:


markbates:
I decided to close this as I don't think it's right for `v0.13` and should remain a target of `v0.14`.

PR #1497, opened 2018-12-09
Issue1064

Shirish111:
Support to run buffalo cli from anywhere inside the buffalo project.


Shirish111:
Will modify the code to avoid Ineffectual Assignment errors

Issue #1512, opened 2018-12-21
Handling browser caching with Buffalo


ycharron:
Buffalo doesn't appear to set (or have a mechanism to set) cache specific HTTP headers when running in production (or other modes). Browser caching is an important aspect of production deployments which provides (significant) performance improvements.

Buffalo has assets fingerprinting to help with caching but no HTTP caching mechanism. Using a proxy in front is not always necessary or feasible.

I can see this being implemented in two ways:
1. Buffalo handles all the request caching itself at a global level (whole app config). Perhaps in app.go.
1. Buffalo provides the ability to override at the request level OR only set at the request level. Although this could be tedious if you have to manually add it on every request. An override seems like a better option.

Those are my thoughts but I'll leave it to the Buffalo devs to implement it the Buffalo way.

These are the 4 headers commonly used for caching:
* ETag
* Cache-Control
* Expires
* Last-Modified

Steps to reproduce:
1. Create a simple Buffalo app
1. Run buffalo build
1. Run in production mode: GO_ENV=production <myapp>
1. Inspect the headers with Chrome or al.

What I expected to see:
* Cache control headers mentioned above.

What I saw:
* No cache headers.

More information on caching:
* https://medium.com/@codebyamir/a-web-developers-guide-to-browser-caching-cc41f3b73e7c

paganotoni:
@ycharron 👋 i know it has been some time since we talked about this, i wanted to clarify something on this issue, this is for asset (JS/CSS/IMAGES) caching, right?

ycharron:
@paganotoni Yes, it's for asset caching. Although it could be extended to support non-asset caching at some point, the primary use case is to cache the assets.

paganotoni:
@ycharron this one already made it to development, it will be out with v0.14.0, thanks for putting together this issue, closing this issue as its already resolved.

PR #1516, opened 2019-01-04
Adding cache control when serving files

paganotoni:
This one is related with #1512 and adds the following cache headers to the file server:

- `ETag` taken from the file mod time (if exists)
- `Cache-Control: max-age=31536000`

cc @ycharron


mclark4386:
Do we want max-age to default to `31536000` but also be a ENV var we can configure?

paganotoni:
Thats a great idea, i'm adding that here.

paganotoni:
@mclark4386 added optional ENV variable

PR #1522, opened 2019-01-04
Fix panic in buffalo info

lukasschlueter:
https://github.com/gobuffalo/meta/commit/0d7e59dd540b2ef77f8dd051c596c8b6f8c660f9 introduced an unexported field to [gobuffalo/meta/app.go](https://github.com/gobuffalo/meta/blob/0d7e59dd540b2ef77f8dd051c596c8b6f8c660f9/app.go#L44), causing a panic in `buffalo info`.

This commit will change the behaviour to ignore inaccessible fields.

### DIFF
--- a/buffalo/cmd/info.go
+++ b/buffalo/cmd/info.go
@@ -31,6 +31,9 @@ var infoCmd = &cobra.Command{
 		var err error
 		for i := 0; i < rt.NumField(); i++ {
 			f := rt.Field(i)
+			if !rv.FieldByName(f.Name).CanInterface() {
+				continue
+			}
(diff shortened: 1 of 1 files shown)


markbates:
@lukasschlueter can you rebase against `master` so we can get this out in a `0.13` patch?

lukasschlueter:
`v0.13.12` is not yet affected (change in `meta` didn't make it there yet).
If we're not planning to add changes (and especially the `meta` update) to the `0.13` versions anymore, I think merging into `development` is the cleaner option as it doesn't involve multiple commits (to master and dev).

But if you prefer, I'll rebase in a few minutes :-)

markbates:
Awesome

PR #1523, opened 2019-01-05
Fingerprinting images

paganotoni:
This PR adds fingerprinted images to our Webpack configuration.

In order to make it backwards compatible it will copy into public both, fingerprinted and plain name files, with this, existing applications can start transitioning to fingerprinted images and we could later on deprecate these (v0.15.x).

Looking forward for your comments on this long awaited feature, thanks for the patience!


markbates:
@paganotoni @stanislas-m what's the status of this? I think I remember some discussion in slack about it, but don't know how it was resolved.

paganotoni:
@markbates duplicate images are there on purpose. This is as way to continue support to 0.13 and prior versions. Another use case for this is supporting needs like serving the image inside an email that's not rendered by buffalo mailers (there is no imgTag like there).

markbates:
I was immediately alarmed that my images were doubling, causing bigger binaries, etc... I can foresee a lot of issues being opened about this, and rightfully so.

This one needs to go back to the drawing board I think. I agree with the fingerprinting, but not so with the duplicate images. We need to rethink this one.

paganotoni:
I don't feel the fundamentals of this one will change (p.e fingerprinting images is a win). IMO we need to get back to the drawing board to think in: 

- how are we going to transition existing users to fingerprinting ? (when they run `buffalo fix`)
- to me this feature undercovers the need for a helper that we can call from actions/libraries to get the path for any fingerprinted asset. 

Overall i think we need to define those 2 before merging this PR, i'm closing it until we define it.

PR #1563, opened 2019-02-03
Overriding app.ErrorMiddleware has no effect fixes #1466

### LINKED ISSUE #1466: Overriding app.ErrorMiddleware has no effect
I have a buffalo.App for a JSON REST api.
Switching the `GO_ENV` to `production` causes all errors to be sent as HTML.
I changed the `app.ErrorMiddleware` to my custom one yet the `defaultErrorMiddleware` is still being called while the one I assigned is ignored.

markbates:
This is a breaking change. We have to replace `app.ErrorMiddleware` with a function, `app.UseErrorMiddleware`. As discussed in #1466 the current `app.ErrorMiddleware` field doesn't have any effect. The reason for this is that the `MiddlewareStack` for the application isn't being updated to replace the original `defaultErrorMiddleware` the app was created with.

This PR adds a new function `app.UseErrorMiddleware` that will update the middleware stack accordingly. It also removes, `app.ErrorMiddleware`, which is a breaking change. 

Seeing as how `app.ErrorMiddleware` has never actually worked, I don't see this affecting too many people.

### DIFF
--- a/app.go
+++ b/app.go
@@ -16,15 +16,14 @@ import (
 type App struct {
 	Options
 	// Middleware returns the current MiddlewareStack for the App/Group.
-	Middleware      *MiddlewareStack `json:"-"`
-	ErrorHandlers   ErrorHandlers    `json:"-"`
-	ErrorMiddleware MiddlewareFunc   `json:"-"`
-	router          *mux.Router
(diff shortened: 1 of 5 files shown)


markbates:
Closing in favor of #1564

PR #1577, opened 2019-02-07
Replace inflect with flect

lukasschlueter:
Note: This also includes the change from #1572 

Several templates are using `opts.Name.Model` where `opts.Name` is an `inflect.Name`, for example: https://github.com/gobuffalo/buffalo/blob/16e4a9e41be133a43c455d98538e3a382b91d3dc/genny/mail/mail.go#L70

I'm not sure what the flect equivalent of this is, so I wasn't able to change the genny option structs yet.


markbates:
Can’t remember off the top of my head as I go to bed (I’m a poet and I don’t even know it), but check out the resource generator, it should have examples of almost all of it.

-----------
Mark Bates

PR #1578, opened 2019-02-07
Make ErrorHandler more robust #1466

sebd71:
Don't expose internal ErrorHandlers map and declare Get/Set methods to manage registered ErrorHandlers.

- make ErrorHandlers Get function more robust (handle nil)
- introduce Clear, Clone, Set and SetMulti functions
- export DefaultErrorHandler function so that user can call it from its
  handler if needed be
- add a new test to check Clear and SetMulti functions

### DIFF
--- a/app.go
+++ b/app.go
@@ -17,7 +17,7 @@ type App struct {
 	Options
 	// Middleware returns the current MiddlewareStack for the App/Group.
 	Middleware    *MiddlewareStack `json:"-"`
-	ErrorHandlers ErrorHandlers    `json:"-"`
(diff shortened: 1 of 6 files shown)


markbates:
Thanks for the feedback. I’m not convinced this needs the full rewrite you propose. The first issue can be fixed with a simple nil check. The second one is expected behavior. If I replace the whole ErrorHandlers with a new one, I would expect it to replace any previous changes I made to it.

I should also point out that you’re requesting breaking changes. I would recommend you open a proposal ticket and lay out your ideas there so the community can discuss the pros/cons of these changes.

PR #1584, opened 2019-02-09
Update SHOULDERS.md

paulwaldmann:
small grammar/spelling fixes


paulwaldmann:
Apologies for the separated PRs, was using the web editor which creates individual branches for each file.

lukasschlueter:
Hi, thanks for the PR!

The `SHOULDERS.md` is automatically created at every release. Changing it directly will only change until the next version update. 

The correct location for this change is the template in https://github.com/gobuffalo/shoulders/blob/master/shoulders/shoulders.go

PR #1604, opened 2019-02-23
Updating bootstrap versions and running packr2 clean

paganotoni:
Updated bootstrap versions and runt `packr2 clean`(@markbates please let me know if that is correct).

cc @stanislas-m


paganotoni:
@markbates merged development here, it will now merge to development, i also noticed it needed to run `pack2` instead of `packr2 clean` (lesson learned).

PR #1639, opened 2019-03-29
Calling error handlers when the url is wrong on the fileserve

paganotoni:
This pr calls the error handler when the fileserver cannot parse the url that's being asked for, it also changes the http code to be 400 given the HTTP specification says:

> The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.

And i feel it would be more adequate for this case.

### DIFF
--- a/buffalo/cmd/root.go
+++ b/buffalo/cmd/root.go
@@ -6,7 +6,7 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


paganotoni:
@markbates added tests for this case and also changed `events.LoadPlugins` to `buffalo.LoadPlugins` because some tests were failing.

PR #1644, opened 2019-04-02
Removing bootstrap 3 support for new apps

paganotoni:


paganotoni:
I thought that given windows was saying something about an invalid escape sequence it may be related with the name of the box, thats why i'm trying twith colons

markbates:
Ah. That’s similar to the error I’m getting. Try using 2.0.6 and see if that helps. I didn’t get a chance to try it.

Convinced it’s something between those versions, but not sure what the problem is. Can’t reproduce it local. Tests pass fine. I’ll fire up the windows box tomorrow and try there.

-----------
Mark Bates

Issue #1642, opened 2019-04-02
Generated font-awesome is out of date


markbates:
We currently generate: 

package.json
`"font-awesome": "~4.7.0"` should be `"@fortawesome/fontawesome-free": "^5.8.1",`

application.scss
`@import "~font-awesome/css/font-awesome.css";` should be `@import "~@fortawesome/fontawesome-free/css/all.min.css";`

jrepe:
Mind if I take this one?

markbates:
Go for it! :)

-----------
Mark Bates

PR #1646, opened 2019-04-05
Generate correct font-awesome for #1642

jrepe:
Changed files according to what was suggested in issue #1642.


markbates:
Thanks!

PR #1655, opened 2019-04-26
Add --module to buffalo build

hdm:
This PR adds the `--module` option to `buffalo build`, which is necessary when building a buffalo application from within a subdirectory of a repository when go modules are enabled.  Without this option, builds from a subdirectory fail with the "cannnot find package" error.


hdm:
Closing this, as the change is necessary, but just one of the many things that break with sub directory builds and modules

PR #1670, opened 2019-05-20
Moving Plugins into Core

mclark4386:
Moving plugins plugin into the core.


mclark4386:
All tests pass locally... its not the cleanest move, but it's a start...

PR #1672, opened 2019-05-24
Add --skip-build-deps option for build

jani123:
When using vendored go modules, one may want to leave out `go get` part from `buffalo build` for various reasons.

My reason to leave module update out is that we have dependencies in private repositories that cannot(or is hard enough to) be fetched while building testing environment for larger set of applications within Docker container.

Other reasons:
* Network connection is down
* You are developing behind expensive mobile network
* There is broken server at GitHub and cannot access repository(has happened 3 times with npm modules for me.. yes, it is unbelievable)
* Dependencies can be saved somewhere and they are saved if they are removed from internet(known to happen) => build still works even though one cannot actually fetch dependencies
* I'm sure there is many other reasons too.


markbates:
Looks like the tests are failing, would you mind taking at look? Thanks.

jani-hast:
> Looks like the tests are failing, would you mind taking at look? Thanks.

Yes, will do asap.

> I was waiting for someone to add this! Thanks!

:) it is good thing  that others can use it too. Clad to help.

jani123:
Okay, so there is actually two bugs in master now that prevents this pull requests pipelines to pass.

Another one is `Test_Template_Partial_Form` test which fails because it uses anonymous struct without ID field. I committed fix for this, but pipelines now finds some different version of some project and form doesn't actually contain action tag as it does on my environment:

```diff
-<form action="/users/user-id" id="user-form" method="POST"><div class="form-group"><label>Name</label><input class=" form-control" id="user-Name" name="Name" type="text" value="Mark" /></div></form>
+<form id="user-form" method="POST"><div class="form-group"><label>Name</label><input class=" form-control" id="user-Name" name="Name" type="text" value="Mark" /></div></form>
```

Another case is that there is this go module support bug which makes tests say:
```
go: unknown subcommand "mod"
```

So the question is that is someone already on those issues since they do happen on other branches too.. I do not have more time for this weekend so, for me this stalls here unfortunately.

markbates:
@jani123 I think this PR should fix the underlying test issues. https://github.com/gobuffalo/buffalo/pull/1673 would love a review if you have a minute.

jani-hast:
@markbates Yes, I inspected the PL through and it shows well who developed Buffalo and who started learning Go and Go Buffalo few weeks ago :D. Looks like that would solve issues. I will merge master here when it is ready.

And on side note, we have in Liana Technologies done now few pull requests with good response and changes has gone through quickly. So, I'd like to thank whole Core Team for that!

markbates:
All passing now! Thanks again!

PR #1675, opened 2019-05-27
Fix skip build deps on cleanup

jani123:
Cleanup was calling `go mod tidy` which started to update existing modules even though I though it is only for cleaning up unused things.


markbates:
Thanks.

PR #1677, opened 2019-05-27
Replacing gometalinter with golangci-lint

paganotoni:
Given Gometaliter is deprecated, this PR moves buffalo to use [golangci-lint](https://github.com/golangci/golangci-lint) as its recommended by Gometalinter deprecation note.


.golangci.yml
```diff
@@ -0,0 +1,14 @@
...
+    - golint
+    - goimports
+    - deadcode
+    - gotype
+    - ineffassign
+    - misspell
+    - nakedret
+    - unconvert
+    - megacheck
+    - varcheck"
```
stanislas-m:
```suggestion
    - varcheck
```

grifts/tools.go
```diff
@@ -18,11 +18,8 @@ var _ = grift.Namespace("tools", func() {
 	}
 
 	var _ = grift.Add("install", func(c *grift.Context) error {
-		if _, err := exec.LookPath("gometalinter"); err != nil {
-			if err := run("go", "get", "-u", "-v", "github.com/alecthomas/gometalinter"); err != nil {
-				return err
-			}
-			if err := run("gometalinter", "--install"); err != nil {
+		if _, err := exec.LookPath("golangci-lint"); err != nil {
+			if err := run("go", "get", "-u", "-v", "github.com/golangci/golangci-lint/cmd/golangci-lint"); err != nil {
```
stanislas-m: Not sure if the update flag is necessary here. :)

Dockerfile.build
```diff
@@ -21,8 +21,7 @@ RUN go get -u github.com/golang/dep/cmd/dep \
 && go get -v -u github.com/markbates/filetest \
 && go get -v -u github.com/markbates/grift \
 && go get -v -u github.com/markbates/refresh \
-&& go get -u github.com/alecthomas/gometalinter \
-&& gometalinter --install
+&& go get -v -u github.com/golangci/golangci-lint/cmd/golangci-lint
```
stanislas-m: Can we use the curl version instead here (since we already have curl in this image)? It will be faster to build and won't download unneeded sources. :)

paganotoni:
@stanislas-m ready to be re-reviewed it

PR #1694, opened 2019-06-13
Decrease docker image size

mrbrownt:
rm commands need to be in the same layer as that adds whatever gets cached or is unneeded. This is because docker layers are append only.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -20,12 +20,12 @@ RUN go get -u github.com/golang/dep/cmd/dep \
 && go get -v -u github.com/gobuffalo/packr/v2/packr2 \
 && go get -v -u github.com/markbates/filetest \
 && go get -v -u github.com/markbates/grift \
-&& go get -v -u github.com/markbates/refresh 
+&& go get -v -u github.com/markbates/refresh \
+&& rm -rfv $GOPATH/src
 
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0
+# Install golangci
+RUN  curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0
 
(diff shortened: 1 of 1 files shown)


markbates:
If I’m reading this correctly you’re deleting all the go source we just downloaded. We do that for a few reasons. Can you explain what this accomplishes?

Dockerfile.build
```diff
@@ -20,12 +20,12 @@ RUN go get -u github.com/golang/dep/cmd/dep \
...
-&& go get -v -u github.com/markbates/refresh 
+&& go get -v -u github.com/markbates/refresh \
+&& rm -rfv $GOPATH/src
 
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0
+# Install golangci
+RUN  curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.16.0
 
-# clear out the src so that we can fill it with new versions
-RUN rm -rfv $GOPATH/src && mkdir -p $BP
```
mrbrownt: I was moving the `rm -rvf` that happens here to the command that's installing all the packages. Everything req'd for buffalo should be downloaded again by the `go get ./...` on line 46 if I'm following along there.

markbates: Got it! It was difficult to make out in the mobile git diff. Thanks.

Dockerfile.build
```diff
@@ -20,12 +20,12 @@ RUN go get -u github.com/golang/dep/cmd/dep \
 && go get -v -u github.com/gobuffalo/packr/v2/packr2 \
 && go get -v -u github.com/markbates/filetest \
 && go get -v -u github.com/markbates/grift \
-&& go get -v -u github.com/markbates/refresh 
+&& go get -v -u github.com/markbates/refresh \
+&& rm -rfv $GOPATH/src
```
mrbrownt: Need to add back in the `&& mkdir -p $BP` I'll push an update shortly

markbates:
Thanks!

PR #1728, opened 2019-07-10
Use model#isPlural to pick which template to auto-render

bytheway:
I ran into an issue attempting to use hyphens instead of underscores in resource URLS (Google recommends underscores not be used: https://support.google.com/webmasters/answer/76329?hl=en).

I traced the limitation down to the autorender, which makes assumptions about the url path segment matching the underscore form. Instead, it is possible to know whether to render the index.html or show.html based on whether we have a collection or a single instance of the model.

### DIFF
--- a/render/auto.go
+++ b/render/auto.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"path"
 	"reflect"
-	"regexp"
 	"strings"
 
 	"github.com/gobuffalo/flect/name"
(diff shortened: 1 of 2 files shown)


markbates:
Can you please add some tests? Thanks. 

Also, Go convention is underscores for file names, which is why Buffalo uses underscores.

bytheway:
Happy to add tests, will do that this evening.

I understand the underscore convention for identifiers and file names (this PR doesn't change that), but that convention gets messy when it runs into the more visible url conventions which have their own style guides, especially when it comes to search engines, etc.

PR #1734, opened 2019-07-13
Add scheme to server address

ruudk:
This makes it easier to CMD + click the URL when the server is started.


stanislas-m:
Thanks! Please note every PR must be opened against the `development` branch, so it can be included in the next release.

ruudk:
Wasn't aware of that, will take that into account in the future :)

PR #1735, opened 2019-07-13
Make the path clickable

ruudk:
When setting up a new project, it's a bit confusing that the path column is empty:
Screenshot 2019-07-13 at 15 51 13

This PR fixes it:
Screenshot 2019-07-13 at 15 51 18


stanislas-m:
Thanks!

Issue #1777, opened 2019-08-29
Conflict between URL and Query String Params 


Max-Pol:
## Description
URL and Query String parameters parsed in the buffalo context — accessible via `c.Params()` — have an unexpected behavior in case of naming conflict, which could lead to a **security issue** in case of URL-based access control.

Basically, only the last parameter set is accessible via `c.Params()`, the others are erased.

### Actual Behavior
Imagine you have a route `GET /users/{user_id}`. If you make the call 
`GET /users/001?user_id=002&user_id=003`, then:
```
c.Params() = map[string][]string {
    "user_id": ["003"]
}
```

If you want the other params (`001, 002`), you need to manually parse the URL... Note that this could be quite a **security issue with URL-based access control**, if the variable set in the URL can be replaced with the query string.

### Possible Expected Behavior
Solution 1. Having the same behavior as the net/url package could be the most consistent :
c.Params() is already a `URL.Values` object, mapping a **string** to a **slice of string**. We just need to append all the parsed params to this last slice, instead of keeping only the last one... Besides, `c.Param("param_name")` could still mimic the `URL.Values.Get` which returns only the **first value** associated with the given key.

Solution 2. Having several ways to retrieve params with the buffalo context:
`c.URLParams()`: `"user_id": ["001"]` in my previous example
`c.QueryStringParams()`:  `"user_id": ["002", "003"]`

### Conclusion
I hope I was clear enough!

markbates:
Why not just use c.Request().URL? It gives you everything you need and doesn’t add more methods. c.Param and c.Params are just quick access helper methods.

Max-Pol:
Sure I agree with you, actually that's what I am currently doing! But that
would be great to continue to use c.Params() which is quite convenient,
instead of having to parse the URL which add some code in every route...
Besides, once again it can imply some security issues with URL-based AC,
for those who are not aware of this unexpected behavior.

I did a small PR that fix it, with a behavior more consistent with
URL.Values. Let me know if you disagree!

Le jeu. 29 août 2019 à 13:21, Mark Bates <notifications@github.com> a
écrit :

PR #1778, opened 2019-08-29
Append URL and Query String parameter dupplicates

Max-Pol:
Fix to issue #1777

A quick fix to be able to manage duplicates in URL Params and query string params.

### Example :
**route**: `GET /users/{user_id}`

**call**: `GET /users/001?user_id=002&user_id=003`

#### Previous Behavior
**`c.Params()`**:
```json
{
    "user_id": [
        "003"
    ]
}
```
**`c.Param("user_id")`**: `"003"`

#### New Behavior
**`c.Params()`**:
```json
{
    "user_id": [
        "001",
        "002",
        "003",
    ]
}
```
**`c.Param("user_id")`**: `"001"`


Max-Pol:
> Can you please add tests for this? Thanks.

I added tests, sorry I should have done it before... 

Thanks for this awesome project btw :)

PR #1780, opened 2019-08-29
use the new buffalo-cli module

markbates:
This PR starts the process of separating the CLI and the Library into two different repos.

This PR uses `type aliases` to allow current  installation instructions and dependencies to migrate to the new repository. In version `v0.16.0` these aliases will be removed and the two projects will be managed separately.


markbates:
Closing for now. #1791

Issue #1781, opened 2019-09-01
buffalo new fails because required git repo is down - how to avoid that?


apimon:
Sorry if I am missing something obvious, I just wanted to take a quick jump into go world, but unfortunately my first attempt to try buffalo did not work, after a very long time of downloading _stuff_ it just stopped:

`go: git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: git fetch -f https://git.apache.org/thrift.git refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/user/go/pkg/mod/cache/vcs/83dba939f95a790e497d565fc4418400145a1a514f955fa052f662d56e920c3e: exit status 128:
	fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out
go: error loading module requirements
`

Seems like that repo is down, ok I get that, however I really would like to know: how to avoid such a problem with a production app? How can I prepare for such outages? 

I have seen that _a lot_ of stuff is downloaded, potentially breaking not only that first step. 

Sorry if this is a go noob question - probably it _is_ because I actually am a go noob... :)

bogh:
I experience the same issue.

robsliwi:
Just tried a small thing: Replacing the import path of the failing thrift go module worked wonders for me.
Maybe someone of you could verify that by checking out https://github.com/gobuffalo/buffalo/pull/1782 and see if this fixes your issue.

markbates:
This isn’t a direct dependency on Buffalo and using replace is only meant for local development. If there is a problem with a transitive dependency the issue should be taken up there.

You may also want to try using a modules proxy to ensure modules are always available.

apimon:
Thanks for the info. For people that where not able to parse the given info into something actionable that help them to actually solve the issue (like me), here is what you could do:

A) set the environment variable `GOPROXY` to some proxy that still delivers the needed dependency, you could do this like so:
```
go env -w GOPROXY=goceinter.io,goproxy.io,goproxy.cn,proxy.golang.org,direct
```
B) the go command uses a Google proxy by default since go 1.13, this version was just released, so you could download that new version and the `buffalo new` command will magically just work, because there is a pre-configured Google proxy that will resolve that dependency:

```
$ go env
...
GOPROXY="https://proxy.golang.org,direct"
```

You do not like these proxies to send data to by default without being asked before?
[...]

PR #1782, opened 2019-09-02
Replace Apache Thrift import for the time being

robsliwi:
This fixes https://github.com/gobuffalo/buffalo/issues/1781


markbates:
This isn’t a direct dependency on Buffalo and using replace is only meant for local development. If there is a problem with a transitive dependency the issue should be taken up there.

PR #1802, opened 2019-10-10
Magic Numbers => HTTP Status Codes

laynemcnish:
Hello!

Long time listener, first time caller.  

I noticed when generating a new Buffalo project that it was using magic numbers (200, 422, 500, etc) instead of using the default status codes from "net/http".  I went ahead and replaced all of the magic numbers with their corresponding status codes and all the tests are passing.

I also snuck in a typo fix for your README.  I can see the irony of fixing that particular sentence while also ignoring its advice and not opening an issue first.

Let me know if you'd like any other information or have any changes you'd like to see and I'd be happy to respond.

Thanks!


laynemcnish:
@stanislas-m I ran packr2, committed the changes and `go test ./...` is successful for me locally.

Issue #1796, opened 2019-10-08
allow passing a timeout to buffalo test


kiambogo:
## Description

Buffalo test should be allowed to take a `timeout` argument, similar `$ go test -timeout 300ms`. 

### Steps to Reproduce the Problem

Buffalo test does not currently provide a timeout option, as seen in the source code (https://github.com/gobuffalo/buffalo/blob/master/buffalo/cmd/test.go) and by testing with the likes of `$ buffalo test -timeout 10m`

### Expected Behavior

Adding a timeout argument should work, allowing the tests to run for the specified duration before it panics. 

### Actual Behavior

Test seems to exit after reloading the schema.

### Info

go 1.13
buffalo v0.14.3

pavan541cs:
@kiambogo Can I work on this issue ?

kiambogo:
Fine by me! I attempted but was having issues with it playing nicely with
other arguments if I recall correctly. Hopefully you can have better luck.

paganotoni:
I confirmed that for some reason this one still not fixed.

paganotoni:
Covered in v0.16.4 thanks @kiambogo

PR #1809, opened 2019-10-18
Add "-timeout" option to "test" command

laynemcnish:
This addresses [this issue](https://github.com/gobuffalo/buffalo/issues/1796) by accepting `-timeout` as an arg to pass to the `mTestRunner` as part of the `pargs`.

I considered renaming `cargs`, `pargs` and `larg` since they were difficult for me, someone new to this codebase, to understand what they stood for.  Thoughts?

Also, I am not sure how to test these commands locally.  Any help with that would be appreciated!


markbates:
Thanks @laynemcnish! Re variable names: Please. Anything that clarifies and helps avoid bugs is always welcome! :)

PR #1811, opened 2019-10-22
Fix test flag bug

ypjama:
This fixes a bug that occurs when --force-migrations flag is used while also specifying the packages to be tested. e.g. `buffalo test ./actions/ --force-migrations`.


ypjama:
Sorry I messed up. I created my branch from master..

Issue #1813, opened 2019-10-23
Sending multiple emails throws error by Google SMTP


jarekjonkisz:
## Description

I would like to be able to send multiple emails using Google SMTP, however I'm being blocked with error message `Google SMTP error 454 4.7.0 Too many login attempts, please try again later.` The error is caused because every email opens and closes a new connection which triggers the above error.

### Steps to Reproduce the Problem

I have tried to send around 100 emails in one go using default Buffalo mailer while I encountered this issue.

### Expected Behavior

I would like to be able to specify when to close the connection.

### Actual Behavior

Connection is closed after each email being sent causing error to be thrown

PR #1815, opened 2019-10-24
Add batch mail sending

Ilkrash:
https://github.com/gobuffalo/buffalo/issues/1813  
This is the proposed solution to properly send a big number of mails at the same time, with opening and closing the connection with SMTP server only one time.


mail/internal/mail/smtp.go
```diff
@@ -199,7 +199,11 @@ func (d *Dialer) DialAndSend(m ...*Message) error {
 	}
 	defer s.Close()
 
-	return Send(s, m...)
+	if errs := Send(s, m...); len(errs) > 0 {
```
mclark4386: as long as there are messages, the length of errs will always be > 0 (example: https://play.golang.org/p/yEQg769VNmz )

Ilkrash: indeed, thank you for the response. Resolved.

Ilkrash:
Please check the fix, tests on Windows failed due to "Test killed: ran too long (10m0s)"

mail/internal/mail/smtp.go
```diff
@@ -199,7 +199,11 @@ func (d *Dialer) DialAndSend(m ...*Message) error {
 	}
 	defer s.Close()
 
-	return Send(s, m...)
+	if errs := Send(s, m...); errs[0] != nil {
```
mclark4386: This is only checking if the first message errored. Please check all message results and send the first error found back.

Ilkrash:
Can I do something to merge these changes? Maybe there are any suggestions to make Batch sending better?

PR #1823, opened 2019-10-31
Updates GolangCI lint

paganotoni:
It seems that GolangCI-lint has some issues with Go 1.13 mentioned [here](https://github.com/golangci/golangci-lint/issues/659). This PR updates the version we're using.


paganotoni:
@stanislas-m can you take a look again please 🙏

Issue #1171, opened 2018-07-15
Logger does not get proper status when calling c.Error


marwan-at-work:
Code to reproduce: 

create a simple buffalo app with a handler that returns `c.Error(500, err)` 

Make sure you pass your own logger to the `App.Logger` field. 

Note that when `WithFields` or `WithField` is called, you never get the `500` status. 

If you do normal `c.Render(status, ...)` then you actually get the appropriate status. 

This is because `c.Error` does not call `c.Response().WriteHeader`. And so, all the middlewares are called first (including the logging middleware), and only after that does the errorHandler call `WriteHeader`

Update: this can also apply to a notFoundHandler, the logger should be able to receive the path params to a route that does not have a handler.

markbates:
Is this  still an issue ?

Max-Pol:
Still an issue, even with the default logger: when `c.Error()` is called, there are actually 2 logs (1 ERROR and 1 INFO) and none of them have a `status`.

paganotoni:
This one was fixed in `v0.16.4` please reopen if still relevant.

PR #1861, opened 2019-12-05
Fix status code & duplicates bug in logs

Max-Pol:
This PR attempts to fix 2 bugs in the logs:
- the status code is missing in the logs when calling `c.Error()` (as describe in the issue #1171 )
- When calling `c.Error()`, two similar logs are produced : 1 that comes from the `request_logger`, 1 from the `defaultErrorHandler`

**Proposal**
- In `c.Error()`: set the `response.Status` used in logs, and add the error message in a logfield `error`
- Only write the logs in request_logger, and set the log level according to the status code

### DIFF
--- a/default_context.go
+++ b/default_context.go
@@ -175,6 +175,12 @@ func (d *DefaultContext) LogFields(values map[string]interface{}) {
 }
 
 func (d *DefaultContext) Error(status int, err error) error {
+	d.LogField("error", err.Error())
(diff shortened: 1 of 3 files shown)


markbates:
@Max-Pol can you please rebase this PR against `development` instead of `master`? Thank you.

https://github.com/gobuffalo/buffalo/blob/master/.github/CONTRIBUTING.md

PR #1870, opened 2019-12-25
Remove Intermediate files

pavan541cs:
This PR aims to remove all the files left over when `buffalo new <app-name>` results in an error.
So incase of any error I am removing the folder itself to remove junk files


pavan541cs:
@stanislas-m Review plz

mclark4386:
I could be wrong, but I thought we left it as is on purpose so as to allow for debugging/fixing manually. Also, there is always `--force` that will delete and remake the folder if it already exists. Would love to here from another contributor on this!

Issue #1839, opened 2019-11-20
new --api  results in not working docker build


yannick:
steps to reproduce:
`go get -u -v -tags sqlite github.com/gobuffalo/buffalo/buffalo` 
`buffalo new myproject --api `
`cd myproject`
`docker build .`
error:
->
`package myproject/actions: unrecognized import path "myproject/actions" (import path does not begin with hostname)`

yannick:
the reason is that the `WORKDIR` in the Dockerfile is set to the name of the directory where `buffalo new` was run in vs. the name of the project.

dschulz:
I just noticed the same problem. It's due to a typo in this line

```
ENV GO111MODULES=on
```

It should be `GO111MODULE` (without the 'S'):   

```
ENV GO111MODULE=on 
```

I tried to find the template or file from which the `Dockerfile` file is generated but I couldn't find any instances of `GO111MODULES` (with an 's') in it in the repo.

dschulz:
I also replaced the line:

```
RUN go get ./...
```

for 
```
RUN go mod tidy
```

The docker image is built without problems after that, but I can't confirm this last change had any effect. The real problem is the typo.

withakay:
The related issue #1854 has been closed but the problem is still present in v15.3.
It looks like the fix was merge to `development` but has not got further than that - https://github.com/gobuffalo/buffalo/pull/1786

PR #1874, opened 2020-01-03
Packing to solve our Dockerfile issue

paganotoni:
This PR packs updated Dockerfile templates to solve #1839. While on it I noticed some tests were failing on Windows because of a GOPATH condition in there. In order to solve that issue, I changed some of the tests to assume modules instead of GOPATH mode.


markbates:
Closing in favor of #1880

PR #1899, opened 2020-02-03
Upgrade to pop/v5 across buffalo

hdm:
This resolves issues with templates and buffalo build using the non-versioned pop package. It does not add additional "fix" substitutions that may be needed.

The specific error this helps with:
>.\buffalo_build_main.go:95:44: cannot use models.DB (type *"github.com/gobuffalo/pop/v5".Connection) as type *"github.com/gobuffalo/pop".Connection in argument to "github.com/gobuffalo/pop".NewMigrationBox


hdm:
This is being reverted, instead all pop/v5 will be moved back to pop for this release.

PR #1900, opened 2020-02-03
Fix indentation

breml:
Replace all spaces with tabs for indentation in templates for Go source code.


breml:
Has this been fixed?

PR #1924, opened 2020-03-18
v0.16.0

markbates:


paganotoni:
Thanks for putting this one together Mark!. The only issue I foresee is that new app tests will not work as `gobuffalo/suite` still not supports `pop/v5`

markbates:
@paganotoni do you have a fix/patch or something?

paganotoni:
There is a branch that changes to use `/v5` but is not blackguard compatible.

paganotoni:
@markbates https://github.com/gobuffalo/suite/tree/pop-v5

paganotoni:
@markbates https://github.com/gobuffalo/suite/pull/36 `gobuffalo/suite/v3` uses pop v5.

buffalo/cmd/fix/fix.go
```diff
@@ -14,14 +14,16 @@ import (
...
 
 var replace = map[string]string{
-	"github.com/markbates/pop":                     "github.com/gobuffalo/pop",
+	"github.com/gobuffalo/buffalo-plugins":         "github.com/gobuffalo/buffalo/plugins",
+	"github.com/gobuffalo/genny":                   "github.com/gobuffalo/genny/v2",
+	"github.com/gobuffalo/pop":                     "github.com/gobuffalo/pop/v5",
+	"github.com/gobuffalo/pop/nulls":               "github.com/gobuffalo/nulls",
+	"github.com/gobuffalo/uuid":                    "github.com/gofrs/uuid",
+	"github.com/markbates/pop":                     "github.com/gobuffalo/pop/v5",
 	"github.com/markbates/validate":                "github.com/gobuffalo/validate",
```
paganotoni: Should Fix move applications to use `validate/v3`?

paganotoni:
@markbates it seems like we need to change https://github.com/gobuffalo/buffalo/blob/76d7d48803d64d66202039d48b7ebda70b580302/genny/build/templates/a/aa.go.plush#L7. Right ?

paganotoni:
Nevermind. It seems you updated that one already :)

paganotoni:
Only thing I noticed is that if we make fix change source to use `uuid/v3`, then we should update validate to use it as well. 
I created this PR to update `Validate` to use `uuid/v3` : https://github.com/gobuffalo/validate/pull/20

paganotoni:
gobuffalo/validate#20 merged @markbates @gobuffalo/all

PR #1926, opened 2020-03-30
fix(render): Windows compatibility for Auto rendering

daHaimi:


render/auto.go
```diff
@@ -113,34 +114,38 @@ func (ir htmlAutoRenderer) Render(w io.Writer, data Data) error {
...
-			return ir.HTML(fmt.Sprintf("%s/new.html", templatePrefix)).Render(w, data)
+			return ir.osIndependentHTML(fmt.Sprintf("%s/new.html", templatePrefix)).Render(w, data)
 		}
 		return nil
 	}
 	cp, ok := data["current_path"].(string)
 
 	defCase := func() error {
-		return ir.HTML(fmt.Sprintf("%s/%s.html", templatePrefix, "index")).Render(w, data)
+		return ir.osIndependentHTML(fmt.Sprintf("%s/%s.html", templatePrefix, "index")).Render(w, data)
```
paganotoni, line 126: @daHaimi could we change this to be 
```suggestion
		return ir.HTML(filepath.Join(tremplatePrefix, "index.html").Render(w, data)
```

And the same across this file?

render/auto.go
```diff
@@ -113,34 +114,38 @@ func (ir htmlAutoRenderer) Render(w io.Writer, data Data) error {
...
 	}
 
 	if !isPlural {
-		return ir.HTML(fmt.Sprintf("%s/show.html", templatePrefix)).Render(w, data)
+		return ir.osIndependentHTML(fmt.Sprintf("%s/show.html", templatePrefix)).Render(w, data)
 	}
 	return defCase()
 }
 
+func (ir htmlAutoRenderer) osIndependentHTML(path string) Renderer {
```
paganotoni, line 145: If we use `filepath.Join` we will not need this method.

daHaimi: I think it is a good idea. Just KISS. :-)

paganotoni:
@daHaimi opened #1951 as an alternative to this one, hoping to release this fix in our next version. Thank you for reporting it!

PR #1933, opened 2020-04-04
Update js template to support font awesome

corylanou:
The latest version of font-awesome requires the js file to also be present for a glyph to work.  To see this work, add this code to any page template:

```html
<h3>
<a href=#" target="_blank"><i class="fab fa-twitter-square share" aria-hidden="true"></i></a>
</h3>
```

Without this JS file, the glyph will not render.

### DIFF
--- a/genny/assets/webpack/templates/assets/js/application.js.tmpl
+++ b/genny/assets/webpack/templates/assets/js/application.js.tmpl
@@ -1,5 +1,6 @@
 require("expose-loader?$!expose-loader?jQuery!jquery");
 require("bootstrap/dist/js/bootstrap.bundle.js");
+require("@fortawesome/fontawesome-free/js/all.js");
 
 $(() => {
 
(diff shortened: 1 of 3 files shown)


genny/assets/webpack/templates/assets/js/application.js.tmpl
```diff
@@ -1,5 +1,6 @@
 require("expose-loader?$!expose-loader?jQuery!jquery");
 require("bootstrap/dist/js/bootstrap.bundle.js");
+require("@fortawesome/fontawesome-free/js/all.js");
```
timraymond, line 3: Should this be `@fontawesome` instead?

corylanou: no for whatever reason they renamed it...  I did a double take the first time as well.  Such is the life of front end development righ?

markbates:
@corylanou can you please rebase against development? thanks.

genny/assets/webpack/templates/assets/js/application.js.tmpl
paganotoni: I confirm what @corylanou says:
https://www.npmjs.com/package/@fortawesome/fontawesome-free

PR #1969, opened 2020-04-16
Add in GZIP compression to file serving

### LINKED ISSUE #1968: Support for gzip compression of static assets
### Description

Compressing static assets like CSS/JS files can lead to big performance improvements. While this is typically done by a frontend proxy like Apache or Nginx, some providers like [Heroku](https://devcenter.heroku.com/articles/compressing-http-messages-with-gzip) do not offer this.

The existing middleware API allows a developer to add compression *except* for files served via `ServeFiles` (static assets) where compression would be most useful. In this case, `ServeFiles` creates a separate `http.Handler` that is outside of the middleware stack.

To solve my immediate problem, I created a quick fix on a [personal branch](https://github.com/vincentchu/buffalo/commit/cf77ce5340724abeb71fb2a3e76073ab74955a17#diff-b1bf5313903dc951eacac638dc5b7060R118) using Gorilla's `CompressionHandler`. I'm happy to clean this up and submit it if the buffalo core team finds it helpful. I'd need some pointers as to the best and cleanest way to get this into buffalo.

As an example, here is a before/after of the size of my served assets (CSS, JS) as measured by pingdom from my production app.  

Screen Shot 2020-04-13 at 12 00 19 PM

vincentchu:
Compressing static assets like CSS/JS files can lead to big performance improvements. While this is typically done by a frontend proxy like Apache or Nginx, some providers like Heroku do not offer this and instead rely on the server to implement content compression.

The existing middleware API allows a developer to add compression *except* for files served via `ServeFiles` (static assets) where compression would be most useful. In this case, `ServeFiles` creates a separate `http.Handler` that is outside of the middleware stack.

To solve this problem for statically-served sites, wrap the file server's `http.Handler` with Gorilla's `CompressionHandler`. This handler just checks the request's `Accept-Encoding` header to enable or disable compression. See: https://godoc.org/github.com/gorilla/handlers#CompressHandler

This change should be invisible to users with the only difference being the return of gzipped responses if the appropriate `Accept-Header` is sent. In my production environment this led to ~2/3rds reduction in bundle sizes. 

/cc @paganotoni 

Fixes #1968

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/gobuffalo/pop/v5 v5.0.11
 	github.com/gobuffalo/tags/v3 v3.1.0
 	github.com/google/go-cmp v0.4.0
-	github.com/gorilla/handlers v1.4.2 // indirect
+	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.7.4
 	github.com/gorilla/sessions v1.2.0
 	github.com/karrick/godirwalk v1.15.5
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/gobuffalo/pop/v5 v5.0.11
 	github.com/gobuffalo/tags/v3 v3.1.0
 	github.com/google/go-cmp v0.4.0
+	github.com/gorilla/handlers v1.4.2 // indirect
 	github.com/gorilla/mux v1.7.4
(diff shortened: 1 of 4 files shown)


vincentchu:
> Totally agree, but like you said, some people still want to handle that using their frontend proxy. :) Can you add an option to disable it?

Done!

paganotoni:
Thanks @vincentchu!

PR #1970, opened 2020-04-16
Fix GO111MODULE value in build Docker images

vladdy:
`ON` is an invalid value for GO111MODULE which prevents Go binaries from executing in the container with error:

```
go: unknown environment setting GO111MODULE=ON
```

This PR fixes the parameter.


vladdy:
@stanislas-m, thank you for the quick turnaround. Please feel free to merge at your convenience as I'm not authorized to do it.

PR #1975, opened 2020-04-20
Binding refactoring

paganotoni:
I started by removing repeated init functions and ended up rewriting the bones of the package. Looking forward to see if this makes sense. My general goal is to make it a bit more testable as well as better separated.


binding/binding.go
```diff
@@ -21,51 +27,26 @@ type Binder func(*http.Request, interface{}) error
...
-	time.RFC1123,
-	time.RFC1123Z,
-	time.RFC3339Nano,
-	time.Kitchen,
-	time.Stamp,
-	time.StampMilli,
-	time.StampMicro,
-	time.StampNano,
+// RequestTypeBinder are those capable of handling a request type like JSON or XML
+type RequestTypeBinder interface {
```
markbates, line 31: This interface feels really, really backward to me. It's a strange level of indirection. As far as I can tell, that is on my struct and its my responsibility to call the `Register` method on the `RequestBinder`. Why?

There should be an interface for a binder.

binding/binding_test.go
```diff
@@ -3,68 +3,46 @@ package binding
...
-	r.Error(err)
+	r.Len(defaultRequestBinder.binders, l+1)
 }
 
-func TestParseTime(t *testing.T) {
+func Test_RegisterCustomDecoder(t *testing.T) {
+	// r := require.New(t)
 
-	r := require.New(t)
+	// RegisterCustomDecoder(func(vals []string) (interface{}, error) {
```
markbates, line 23: lots of commented out code kicking about

binding/file_request_type_binder.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/monoculum/formam"
+)
+
+// FileRequestTypeBinder is in charge of binding File request types.
+type FileRequestTypeBinder struct{}
+
+// RegisterTo register this RequestTypeBinder to the passed request binder
+// on the File content types (multipart/form-data).
+func (ht FileRequestTypeBinder) RegisterTo(binder *RequestBinder) {
+	binder.Register("multipart/form-data", ht.binder(binder.formDecoder))
```
markbates, line 16: `binder.formDecoder` is not-exported, so how would a third party write this same binder?

binding/request_binder.go
```diff
@@ -0,0 +1,111 @@
...
+
+var (
+	errBlankContentType = errors.New("blank content type")
+)
+
+// RequestBinder is in charge of binding multiple requests types to
+// structs.
+type RequestBinder struct {
+	lock    *sync.Mutex
+	binders map[string]Binder
```
markbates, line 24: consider using a `sync.Map`? You can use https://github.com/gobuffalo/mapgen to generate a typed version of the map. its a little cleaner to use than  the mutex.

binding/request_binder.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+var (
+	errBlankContentType = errors.New("blank content type")
+)
+
+// RequestBinder is in charge of binding multiple requests types to
+// structs.
+type RequestBinder struct {
+	lock    *sync.Mutex
```
markbates, line 23: if using a mutex here, consider a `RWMutex`

binding/request_binder.go
```diff
@@ -0,0 +1,111 @@
...
+	if ba, ok := value.(Bindable); ok {
+		return ba.Bind(req)
+	}
+
+	ct := httpx.ContentType(req)
+	if ct == "" {
+		return errBlankContentType
+	}
+
+	if b, ok := rb.binders[ct]; ok {
```
markbates, line 58: You ned to wrap this with a lock.

binding/binding.go
paganotoni: Thanks @markbates!

binding/request_binder.go
paganotoni: Thanks Mark. Added.

binding/request_binder.go
paganotoni: Done

binding/binding.go
paganotoni: Covered this one.

binding/file_request_type_binder.go
paganotoni: I did rewrite it. Take a look at it and LMK if it makes sense.

binding/request_binder.go
paganotoni: Did not get this one. Will ask you offline.

binding/binding_test.go
paganotoni: This one still pending I'll cover it soon.

Issue #1989, opened 2020-05-12
Bootstrap compilation requires autoprefixer - not included in webpack today


arunsworld:
### Description

Compiling bootstrap from sass as part of webpack build pipeline requires autoprefixer to work correctly (https://getbootstrap.com/docs/4.0/getting-started/build-tools/)

However, this is not included in the default webpack config.

I'm very happy to submit a PR for this if this issue is accepted.

### Steps to Reproduce the Problem

1. Setup a buffalo application with webpack enabled
2. In index.plush.html add the following lines of HTML somewhere in the body: `<input type="range" class="custom-range" id="customRange1">`
3. Note that the range is not shown correctly like it should (https://getbootstrap.com/docs/4.1/components/forms/#range)

### Expected Behavior

Custom range should be rendered as per documentation.

### Actual Behavior

Actual rendering is unusable because vendor prefixes are missing.

The class custom-range is missing vendor prefix (eg. -webkit-appearance:none).

To fix this we need to have postcss-loader in webpack config with autoprefixer plugin.

Happy to provide PR.

### Info

Please run `buffalo info` and paste the information below where it says "PASTE_HERE".

<details>

```
-> Go: Checking installation
✓ The `go` executable was found on your system at: /usr/local/go/bin/go

-> Go: Checking minimum version requirements
✓ Your version of Go, 1.14.2, meets the minimum requirements.

-> Go: Checking Package Management
✓ You are using Go Modules (`go`) for package management.

-> Go: Checking PATH
✓ Your PATH contains /Users/abarua/go/bin.

-> Node: Checking installation
✓ The `node` executable was found on your system at: /usr/local/bin/node

-> Node: Checking minimum version requirements
✓ Your version of Node, v12.16.3, meets the minimum requirements.

-> NPM: Checking installation
✓ The `npm` executable was found on your system at: /usr/local/bin/npm

-> NPM: Checking minimum version requirements
✓ Your version of NPM, 6.14.4, meets the minimum requirements.

-> Yarn: Checking installation
✓ The `yarnpkg` executable was found on your system at: /usr/local/bin/yarnpkg

-> Yarn: Checking minimum version requirements
✓ Your version of Yarn, 1.22.4, meets the minimum requirements.

-> PostgreSQL: Checking installation
✘ The `postgres` executable could not be found on your system.
For help setting up your Postgres environment please follow the instructions for you platform at:

https://www.postgresql.org/download/

-> MySQL: Checking installation
[...]

PR #1990, opened 2020-05-18
added autoprefixer for bootstrap compilation

arunsworld:
Modifies webpack.config.js to include postcss-loader which is setup to use autofixer when processing css files. Fixes issue: #1989.

Tests have been accordingly modified.


paganotoni:
Thanks @arunsworld

PR #1999, opened 2020-05-28
Bump spf13/viper to v1.7.0

aeneasr:
Version 1.7.0 includes security related bugfixes, including a published CVE for gorilla/websocket which is an indirect dependency of spf13/viper v1.7.0 and can cause issues in downstream repositories when below 1.4.0 (which is the case for the currently pinned version).


aeneasr:
Thank you for the quick merge!

PR #2008, opened 2020-06-11
GitHub Actions upgrade to actions/checkout@v2

cclauss:
https://github.com/actions/checkout/releases

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,8 +8,8 @@ jobs:
       matrix:
         go-version: [1.13.x, 1.14.x]
         os: [macos-latest, windows-latest, ubuntu-latest]
-    env:
-      GO111MODULE: on
+    # env:
+    #  GO111MODULE: on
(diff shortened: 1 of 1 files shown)


cclauss:
Not making progress...

psadac:
I think I have the solution for your problem. You need to add a `setup-go@v2` action in your workflow with the selected Go version : 

    - name: Install Go
      uses: actions/setup-go@v2
      with:
        go-version: ${{ matrix.go-version }}

If you don't, all the tests will be run with an arbitrary version.
Moreover, you can use  `checkout@v2` action without precising the repository or the depth : 

    - name: Checkout code
      uses: actions/checkout@v2
      with:
        submodules: true

for the complete file `test.yml`

https://github.com/psadac/buffalo/blob/master/.github/workflows/tests.yml

All the tests seems to pass on my forked repository :  

https://github.com/psadac/buffalo/actions/runs/133798914

hope it helps.

Issue #2105, opened 2021-05-14
Content-Type not correctly set in error response headers


StefanDorresteijn:
### Description

Whenever a `c.Error` is called from an action, the content-type header in the response is set to `text/plain; charset=utf-8`. This happens on a custom error handler (where I call `res.Header().Set("Content-Type", "application/json")` directly) as well as the default handler. 

This seems to happen regardless of the content-type set in the request. While the rendered content changes, the header does not.

I think this has to do with a `res.WriteHeader` call before calling `res.Header().Set()`, locking the headers, because no other header values can be set either. Once the `res.WriteHeader` call is moved after the `res.Header().Set()` calls, it seems to work again.

### Steps to Reproduce the Problem

  Please describe in painful detail what you did (so others can play along with you) to get to this point. This includes things like the exact command(s) you used, or the `curl` command you used, that sort of thing.

  1. Create a route
  2. Return `c.Error` from that route

### Expected Behavior

I'd expect to have the `Content-Type` header set to the request's `Content-Type` header, or if it's set specifically in the error handler to not be overwritten.

### Actual Behavior

The response header is always set to `text/plain; charset=utf-8`.

### Info

Please run `buffalo info` and paste the information below where it says "PASTE_HERE".

<details>

```
stefan@Stefans-MBP endpoint_manager % buffalo info
-> Go: Checking installation
✓ The `go` executable was found on your system at: /opt/homebrew/homebrew/bin/go

-> Go: Checking minimum version requirements
✓ Your version of Go, 1.14.15, meets the minimum requirements.

-> Go: Checking Package Management
✓ You are using Go Modules (`go`) for package management.

-> Go: Checking PATH
✓ Your PATH contains /Users/stefan/go/bin.

-> Node: Checking installation
✓ The `node` executable was found on your system at: /opt/homebrew/homebrew/bin/node

-> Node: Checking minimum version requirements
✓ Your version of Node, v15.12.0, meets the minimum requirements.

-> NPM: Checking installation
✓ The `npm` executable was found on your system at: /usr/local/bin/npm

-> NPM: Checking minimum version requirements
✓ Your version of NPM, 7.6.0, meets the minimum requirements.

-> Yarn: Checking installation
✓ The `yarnpkg` executable was found on your system at: /usr/local/bin/yarnpkg

-> Yarn: Checking minimum version requirements
✓ Your version of Yarn, 1.22.10, meets the minimum requirements.

-> PostgreSQL: Checking installation
[...]

PR #2106, opened 2021-05-14
Write status header after setting all other headers

StefanDorresteijn:
The status header was set before `content-type`, which locked the header, making all other calls to `Header().Set()` fail. This fixes https://github.com/gobuffalo/buffalo/issues/2105


StefanDorresteijn:
I've made a few changes to clean up the code but I'm not sure it's exactly how it should be implemented. I'm just adding a function to the `DefaultContext` struct now, and I'm not yet convinced that that function should be there. However, without that function I'm not sure how I'd clean up the code.

fasmat:
Thanks for your contribution!

Could you inline the method `writeHeaders`? It is only defined for the `DefaultContext` so the new code would panic if a custom `buffalo.Context` is used for `c` instead.

Issue #2074, opened 2021-01-05
build fails with golang 1.16.0


giulianozor:
Hello,
I am trying to build by project with golang ~1.16beta1~ 1.16.0 (using the alpine based docker image) and I get the following error:
```
buffalo build -v
DEBU[2021-01-05T07:39:43Z] Step: e1807c46
DEBU[2021-01-05T07:39:43Z] Chdir: /test/xxx
DEBU[2021-01-05T07:39:43Z] File: /test/xxx/main.go
DEBU[2021-01-05T07:39:43Z] File: /test/xxx/a/aa.go
DEBU[2021-01-05T07:39:43Z] File: /test/xxx/buffalo_build_main.go
DEBU[2021-01-05T07:39:43Z] File: /test/xxx/migrations/.pop-tmp.md
DEBU[2021-01-05T07:39:43Z] Exec: go build -v -tags development -o bin/xxx
go: updates to go.mod needed; try 'go mod tidy' first
DEBU[2021-01-05T07:39:43Z] Rollback: main.go
DEBU[2021-01-05T07:39:43Z] File: /test/xxx/main.go
DEBU[2021-01-05T07:39:43Z] Delete: a/aa.go
DEBU[2021-01-05T07:39:43Z] Delete: buffalo_build_main.go
DEBU[2021-01-05T07:39:43Z] Delete: migrations/.pop-tmp.md
DEBU[2021-01-05T07:39:43Z] Exec: go mod tidy
Usage:
  buffalo build [flags]

Aliases:
  build, b, bill, install

Flags:
      --clean-assets               will delete public/assets before calling webpack
      --dry-run                    runs the build 'dry'
      --environment string         set the environment for the binary (default "development")
  -e, --extract-assets             extract the assets and put them in a distinct archive
  -h, --help                       help for build
      --ldflags string             set any ldflags to be passed to the go build
      --mod string                 -mod flag for go build
  -o, --output string              set the name of the binary
  -k, --skip-assets                skip running webpack and building assets
      --skip-build-deps            skip building dependencies
      --skip-template-validation   skip validating templates
  -s, --static                     build a static binary using  --ldflags '-linkmode external -extldflags "-static"'
  -t, --tags string                compile with specific build tags
  -v, --verbose                    print debugging information

ERRO[0000] Error: exit status 1
```

A strange thing is that echoing $? returns 255, not 1

Running ```go mod tidy``` does not solve the issue.
Running ```buffalo dev``` works fine
[...]

PR #2113, opened 2021-06-15
Fix build issues with golang 1.16

fasmat:
Hi,

this should fix the issue mentioned in #2074. The generated `a` package should not be cleaned up before go mod tidy is being run, but rather the other way round.

I also updated some dependencies of gobuffalo to their most recent versions.


go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/gobuffalo/buffalo
 
-go 1.15
+go 1.16
```
paganotoni, line 3: @fasmat Am I right on thinking changing the version on the Go.mod to v1.16 will make buffalo compatible only with Go 1.16+ ? If so I think we should wait for v1.17 to be out to make this chance.

paganotoni (commented):
Left a question in there, besides that it looks good to me. One we address that question we can see if we can move forward. thanks @fasmat

go.mod
fasmat: Yes, the version statement indicates to go which is the oldest version of go that the module is compatible with (since newer versions will compile it anyway). I will revert it back to 1.15.

PR #2128, opened 2021-07-19
Update dependencies after buffalo new

fasmat:
This PR adds a new step to ``buffalo new``:
All dependencies listed in `go.mod` are upgraded. Upgrades are only performed on direct dependencies to avoid any issues that can occur when updating transitive dependencies.


fasmat:
Any feedback on this PR @paganotoni ?

PR #2146, opened 2021-09-11
Improve performance of WrapBuffaloHandler

fasmat:
This PR adds performance improvements to the WrapBuffaloHandler function.

In one of my apps I noticed that WrapBuffaloHandler is very slow for what it is trying to do. The reason is that every time the wrapped handler is called a new instance of a buffalo.App is created which is very slow and unnecessary.

This PR should fix the problem. The first commit only contains a benchmark and the commit(s) after that the changes, to be able to see the differences. On my machine I was able to achieve a ~ 40x performance increase with this change.

I also extended the documentation of the relevant function to inform the user that the functionality comes with some drawbacks.


fasmat:
I simplified the code by merging the functionality of the wrappedContext (only used in wrapped handlers) into the default context.

stanislas-m (approved):

paganotoni:
Thanks @fasmat  and @stanislas-m !

PR #2180, opened 2021-11-27
Add reload.FS for hot reloading during development

fasmat:
This adds an implementation of `fs.FS` that can do both: hot reload and embedding of files. `embed.go` is automatically hidden from the embedded file system.

There is a related PR here: https://github.com/gobuffalo/cli/pull/60

This PR updates the new command to generate `embed.go` files that use the `reload.FS`.


reload/fs.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (f FS) Open(name string) (fs.File, error) {
+	if name == "embed.go" {
+		return nil, fs.ErrNotExist
+	}
+	file, err := f.getFile(name)
+	if name != "." {
+		return file, err
+	}
+	return rootFile{file}, err
```
sio4, line 29: It could be easier to read if this block is

```go
if name == "." {
    return rootFile{file}, err
}
return file, err
```

reload/fs.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (f FS) getFile(name string) (fs.File, error) {
+	cfgFile := "./.buffalo.dev.yml"
+	if _, err := os.Stat(cfgFile); err != nil {
+		return f.embed.Open(name)
+	}
+	return f.dir.Open(name)
+}
```
sio4, line 38: This is a design decision. How about allowing live access even in production mode outside the application tree? E.g. just try `f.dir.Open` first and do `f.embed.Open` when it failed? What do you think @paganotoni ?

When I developed a small in-house application with buffalo about 4 years ago, it was very useful when I need to update some HTML files, updated images, and auto-response templates without code change.

reload/fs.go
```diff
@@ -0,0 +1,60 @@
+package reload
```
sio4, line 1: How about renaming it as something like "overlay", "live", "shadow", or... just simply "buff", something closer to what it does? Actually, it does not "reload" but act as a kind of "overlay". "reload" is fine but I would like to make it clear to understand when the user uses it.

reload/fs.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+type rootFile struct {
+	fs.File
+}
+
+func (f rootFile) ReadDir(n int) ([]fs.DirEntry, error) {
+	dir, ok := f.File.(fs.ReadDirFile)
+	if !ok {
+		return nil, fmt.Errorf("failed at hiding embed.go file")
```
sio4, line 47: Is it a correct error message? It could be "called on incompatible object" or something.

reload/fs.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	entries, err := dir.ReadDir(n)
+	if err != nil {
+		return entries, err
+	}
+
+	for i, entry := range entries {
+		if entry.Name() == "embed.go" {
+			entries = append(entries[:i], entries[i+1:]...)
+		}
```
sio4, line 57: It is buggy that if "embed.go" is there, we will return n-1 entries.
If we cannot do this when `NewFS()` is called and construct the object, maybe we need to do `dir.ReadDir(n+1)` and drop the last entry by a certain condition... What do you think?

sio4 (changes requested):
Hi @fasmat, nice work!
I added some comments and suggestions. Please consider them.

sio4:
I tested creating a new app with this PR along with gobuffalo/cli#60 and the result was excellent! the live loading worked perfectly as before!

I think we can merge it once we agreed with the name of the module, and some minor bugs could be managed after this hotfix was released.

reload/fs.go
fasmat: I now changed the implementation to try `f.dir.Open` and only when it fails fall back to `f.embed.Open`. This would then also require the built binary to be run from the same folder as `main.go` for this hot swapping in a deployed environment to work.

I personally hav only ever used a buffalo application within a docker image. So when we update assets we would need to build a new image anyway and have it deployed.

reload/fs.go
fasmat: I replaced it with an error that `f` isn't a directory.

reload/fs.go
sio4: Yeah, I agree that is a modern and standard way to deploy services nowadays.

reload/fs.go
fasmat: It's not really a bug. The documentation to `fs.ReadDirFile` ([here](https://pkg.go.dev/io/fs#ReadDirFile)) says:

> If n > 0, ReadDir returns **at most** n DirEntry structures.
> In this case, if ReadDir returns an empty slice, it will return
> a non-nil error explaining why.
> At the end of a directory, the error is io.EOF.

> If n <= 0, ReadDir returns all the DirEntry values from the directory
> in a single slice. In this case, if ReadDir succeeds (reads all the way
> to the end of the directory), it returns the slice and a nil error.
> If it encounters an error before the end of the directory,
> ReadDir returns the DirEntry list read until that point and a non-nil error.

So the current implementation would have been fine. I still changed it, so when `n > 0` it reads `n + 1` Entries from the underlying FS and then remove `embed.go`. If for some reason `embed.go` wasn't found the last entry is removed instead to return `n` entries.

fasmat:
> I tested creating a new app with this PR along with [gobuffalo/cli#60](https://github.com/gobuffalo/cli/pull/60) and the result was excellent! the live loading worked perfectly as before!
> 
> I think we can merge it once we agreed with the name of the module, and some minor bugs could be managed after this hotfix was released.

How about we move this functionality to the root of the `gobuffalo/buffalo` project? Then the name would be `buffalo.FS` which seems fitting to me?

reload/fs.go
fasmat: How about we move this functionality to the root of the gobuffalo/buffalo project? Then the name would be buffalo.FS which seems fitting to me?

reload/fs.go
sio4: My interpretation of **at most n** is that sometimes it could return `m` (`m`<`n`) entries if there are only `m`, fewer files than `n`. So if it just returns 5 when I call with 6, the user could consider it as there are only 5 files in the directory. I don't think this is critical for our usage but I just want to be clear with it :-)

Anyway, Thanks for your kind consideration!

sio4:
> How about we move this functionality to the root of the `gobuffalo/buffalo` project? Then the name would be `buffalo.FS` which seems fitting to me?

Wow! I love the name!

sio4 (approved):
Nice work! I hope we can have it as soon as possible!

paganotoni:
Hey Folks! This is exciting, as for the hot reloading mechanics, just to confirm:

- while in development mode GO_ENV=development this will use a Dir FS.
- while in production mode GO_ENV=production this will use the embedded FS

Am I right with those assumptions?

paganotoni:
@fasmat I love it! It's really exciting to see this progress! Adding here my 5 cents:

I would suggest to add some tests to `buffalo.FS` as it will be an important (and very much used) element in gobuffalo/* and users codebases. 

Also, and to add some context, there was a security issue in Packr we should keep in mind here. At some point Packr was allowing to read relative folders, opening the entire docker container to the world. This was a security issue as things like environment variables usually are stored in OSs files.

As for how this should work we need to make our buffalo.FS consider the environment variable (GO_ENV) and only use disk FS when `development` or `test`. Otherwise always use the embedded folder.

fasmat:
> Hey Folks! This is exciting, as for the hot reloading mechanics, just to confirm:
> 
> * while in development mode GO_ENV=development this will use a Dir FS.
> * while in production mode GO_ENV=production this will use the embedded FS
> 
> Am I right with those assumptions?

No this is not true. It behaves similarly to how packr functioned: if the requested files can be found on the disk, those will be used. Otherwise it will use the embedded files.

fasmat:
> @fasmat I love it! It's really exciting to see this progress! Adding here my 5 cents:
> 
> I would suggest to add some tests to `buffalo.FS` as it will be an important (and very much used) element in gobuffalo/* and users codebases.

@paganotoni Thanks for the feedback 😄  I'll see what I can do. The functionality is very limited. The whole thing is just 3 methods that essentially just override the functions of embed.FS and os.DirFS to hide `embed.go`.

> Also, and to add some context, there was a security issue in Packr we should keep in mind here. At some point Packr was allowing to read relative folders, opening the entire docker container to the world. This was a security issue as things like environment variables usually are stored in OSs files.
> 
> As for how this should work we need to make our buffalo.FS consider the environment variable (GO_ENV) and only use disk FS when `development` or `test`. Otherwise always use the embedded folder.

I don't think this is necessary @sio4 mentioned the use case of not having embedded files in production, which used to work with packr and maybe some other users also rely on. Regarding reading relative folders: both `embed.FS` and `os.DirFS` don't allow this. There is no way that `buffalo.FS` would introduce this bug by just wrapping those two.

paganotoni:
ok. It's great to know that stdlib is covering that issue. I would suggest that as part of testing the fallback behavior we're implementing in buffalo.FS we add some of those relative paths test, maybe just one to have as a guarantee of what we expect from this FS implementation.

paganotoni (approved):
Awesome. Thanks @fasmat !

PR #2186, opened 2021-12-09
Buffalo should respect http status when returning error

dmuriel:
When returning an HTTP Status different than 404 caused by an "sql.ErrNoRows" buffalo is automatically changing the error status to a 404 error. Buffalo should respect the error status passed on the return.

### DIFF
--- a/route_info_test.go
+++ b/route_info_test.go
@@ -2,12 +2,12 @@ package buffalo
 
 import (
 	"database/sql"
+	"fmt"
 	"net/http"
 	"testing"
 
 	"github.com/gobuffalo/buffalo/render"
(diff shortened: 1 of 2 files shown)


route_info_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/gobuffalo/buffalo/render"
 	"github.com/gobuffalo/httptest"
+	"github.com/pkg/errors"
```
paganotoni, line 10: @dmuriel I suggest here using fmt.Errorf("msg: %w" to avoid the external dependency.

paganotoni (approved):
Thanks @dmuriel

Issue #2112, opened 2021-06-04
Session Cookie set twice in response headers


saurori:
### Description

It looks like the Session Cookie is set twice in the response headers, and it is not the same value.

### Steps to Reproduce the Problem

```bash
$ buffalo new coke
...
$ buffalo dev
...
```

```bash
$ curl -i localhost:3000             
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Set-Cookie: _coke_session=MTYyMjg0ODIzOHxEdi1CQkFFQ180SUFBUkFCRUFBQU92LUNBQUVHYzNSeWFXNW5EQTRBREhKbGNYVmxjM1J2Y2w5cFpBWnpkSEpwYm1jTUZnQVVOVEEyWkdWaFlUazJPVGRrWVdKaU5UUXhPVGs9fH-MqWSmwpvvEGNKRnJBmQs1hgFfkOkbHfYtow7k8GAw; Path=/; Expires=Sun, 04 Jul 2021 23:10:38 GMT; Max-Age=2592000; HttpOnly
Set-Cookie: _coke_session=MTYyMjg0ODIzOHxEdi1CQkFFQ180SUFBUkFCRUFBQV82UF9nZ0FEQm5OMGNtbHVad3dPQUF4eVpYRjFaWE4wYjNKZmFXUUdjM1J5YVc1bkRCWUFGRFV3Tm1SbFlXRTVOamszWkdGaVlqVTBNVGs1Qm5OMGNtbHVad3dVQUJKaGRYUm9aVzUwYVdOcGRIbGZkRzlyWlc0SFcxMTFhVzUwT0FvaUFDRGRUdW1HY0V3b3FicVVISlpWVzhSRTIybWJjUUlSeEJDNkQyN1dqbWZYdUFaemRISnBibWNNQ1FBSFgyWnNZWE5vWHdkYlhYVnBiblE0Q2dRQUFudDl8s8JVDn8xNHPJmBatfG1XrRV993BtGnvEAWhaSGsUbRI=; Path=/; Expires=Sun, 04 Jul 2021 23:10:38 GMT; Max-Age=2592000; HttpOnly
Date: Fri, 04 Jun 2021 23:10:38 GMT
Transfer-Encoding: chunked
```

### Expected Behavior

The response headers set the Session Cookie once (one header).

### Actual Behavior

The response has two `Set-Cookie` headers for the Session.

### Info

- `Buffalo version is: v0.16.23`
[...]

PR #2193, opened 2021-12-17
Only save Session once to avoid duplicate session cookie

saurori:
Addresses this Issue: https://github.com/gobuffalo/buffalo/issues/2112

`Session().Save()` was being called multiple times. If the session data changes between calls to `Save()`, the session cookie value would be different and therefore set multiple times. `Session().Save()` was also called in flash `persist()` and was therefore coupled tightly to Flash functionality.

Session saving is now done only once in `Render()` or `Redirect()`. There is still the possibility of a user calling `c.Session().Save()` in a request handler or similar and cause a duplicate session cookie to write.

### DIFF
--- a/route_info.go
+++ b/route_info.go
@@ -103,7 +103,6 @@ func (ri RouteInfo) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 
 	events.EmitPayload(EvtRouteStarted, payload)
 	err := a.Middleware.handler(ri)(c)
-	c.Flash().persist(c.Session())
 
 	if err != nil {
 		status := http.StatusInternalServerError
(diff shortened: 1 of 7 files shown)


route_info.go
```diff
@@ -103,7 +103,6 @@ func (ri RouteInfo) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 
 	events.EmitPayload(EvtRouteStarted, payload)
 	err := a.Middleware.handler(ri)(c)
-	c.Flash().persist(c.Session())
```
saurori, line 106: It was not clear why this is needed, so I removed it. `persist()` no longer saves Session, so is this needed? Only reference is: https://github.com/gobuffalo/buffalo/pull/2054

saurori:
Commenting to remove stale. Still relevant, waiting for review 😢

fasmat (approved):

PR #2206, opened 2022-02-03
Configure Dependabot to merge to development instead of main

fasmat:


paganotoni (approved):
Love it. Thanks @fasmat !

PR #2226, opened 2022-03-24
Fixed multi-homing feature and related bugs. (#2222)

### LINKED ISSUE #2222: "Feature/Addition Host domain" #2185 was a breaking change.
### Description

Oops! #2185 was a breaking change.

Previously, also currently, `App` embeds `Options` and `Options` has `Host`. So till v1.18.1, we can access configured Host by `app.Host` and users could use it e.g. https://gobuffalo.io/en/docs/goth/.

```
type App struct {
	Options
	<...>
```

```
type Options struct {
	Name string `json:"name"`
	Addr string `json:"addr"`
	Host string `json:"host"`
```

By this change, apps referenced the value by `app.Host` will fail to get the hostname. (Actually, my project used that.)

The workaround is to change applications by replacing `app.Host` with `app.Options.Host` but I don't think this is a good thing since we also allow users to access the other values with the same convention like `app.Name`.

I will check how the change works and what could be a better fix. (considering both of changing the new function's name or add a new function to get the value (including multi-homed Host) or just let it as is.)

It seems like this is more likely a design issue by the way. Please provide any ideas or opinions.

sio4:
This will fix #2222.

In brief,
- added structure `Home` as an interim solution (hope to be reshaped as `Group` structure in v1)
- adapted to the new structure
- fixed bugs related to this feature and pre-existing small ambiguous things (related to routes and middleware handling)
- renamed the function from `Host()` to `VirtualHost()` to make the name clear, and avoid conflicting with `App.Options.Host`.
- added clarified function descriptions for involved functions.

More details (including the details of fixed bugs or symptoms) will be added on issue #2222 to consolidate the information/history.

fixes #2222

### DIFF
--- a/app.go
+++ b/app.go
@@ -58,6 +58,8 @@ func New(opts Options) *App {
 
 		RouteNamer: baseRouteNamer{},
 	}
+	a.Home.app = a     // replace root.
+	a.Home.appSelf = a // temporary, reverse reference to the group app.
 
 	dem := a.defaultErrorMiddleware
 	a.Middleware = newMiddlewareStack(dem)
--- a/app.go
+++ b/app.go
@@ -14,15 +14,13 @@ import (
 // Without an App you can't do much!
 type App struct {
 	Options
-	// Middleware returns the current MiddlewareStack for the App/Group.
-	Middleware    *MiddlewareStack `json:"-"`
-	ErrorHandlers ErrorHandlers    `json:"-"`
-	router        *mux.Router
-	moot          *sync.RWMutex
-	routes        RouteList
-	root          *App
-	children      []*App
-	filepaths     []string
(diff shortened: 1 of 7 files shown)


sio4:
oops? why CI is not running?

dmuriel:
Hey @sio4, just wanted to thank you. I appreciate the work you did in this PR.

paganotoni:
This looks good to me @sio4. I'm not sure how to run the tests against these (We're in github limbo). I wonder if you could send this to a branch within the gobuffalo/gobuffalo repo to see if that triggers the changes. Appreciate this @sio4!

paganotoni (commented):
Merging this. We can test it in development and release depending on how it behaves.

sio4:
Thanks! I lately found your comment. :-)

PR #2233, opened 2022-04-06
Update quick start links as they're 404ing

aporcupine:
👋  Noticed when taking a look at buffalo that the quick start links are 404ing 🙈 

Figured I'd update these as it's one of the first places a new user will click.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ Please visit [http://gobuffalo.io](http://gobuffalo.io) for the latest documenta
 
 ### Quick Start
 
-- [Installation](http://gobuffalo.io/docs/installation)
-- [Create a new project](http://gobuffalo.io/docs/new-project)
-- [Examples](http://gobuffalo.io/docs/examples)
+- [Installation](https://gobuffalo.io/documentation/getting_started/installation)
+- [Create a new project](https://gobuffalo.io/documentation/getting_started/new-project)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,9 +34,9 @@ Please visit [http://gobuffalo.io](http://gobuffalo.io) for the latest documenta
 
 ### Quick Start
 
-- [Installation](http://gobuffalo.io/docs/installation)
-- [Create a new project](http://gobuffalo.io/docs/new-project)
-- [Examples](http://gobuffalo.io/docs/examples)
+- [Installation](https://gobuffalo.io/documentation/getting_started/installation)
+- [Create a new project](https://gobuffalo.io/documentation/getting_started/new-project)
+- [Tutorials](https://gobuffalo.io/documentation/tutorials/)
```
aporcupine, line 39: Looks like the title of this page has changes, so I updated the link name.

paganotoni: Thanks!!!

sio4:
@paganotoni, it seems like the codeclimate integration is not working. Do we still use that? It seems like the button for the climate is not on the README.

paganotoni:
I've been wanting to move to something else. But thats another topic we can talk about. I think we could benefit from something like golangci-lint. Maybe food for thought.

PR #2185, opened 2021-12-07
Feature/Addition Host domain

dmuriel:
This PR adds new functionality for Buffalo, by using the [Host](https://github.com/gorilla/mux#matching-routes) function from Gorilla/Mux, buffalo can now restrict the handlers and render those that only match the specified host. This is very useful if you have multiple portals in the same application and/or want to use the same path for different portals.

```
...

func HomeApp(c buffalo.Context) error {
	return c.Render(http.StatusOK, r.String("OK APP"))
}

func HomeLMS(c buffalo.Context) error {
	return c.Render(http.StatusOK, r.String("OK LMS"))
}

func New() *buffalo.App {
	bapp := buffalo.New(buffalo.Options{
		Env:         envy.Get("GO_ENV", "development"),
		SessionName: "_application_session",
	})

        app := bapp.Host("app.domain.com")
        app.GET("/dashboard", HomeApp) // This will render the dashboard for the Application

        lms := bapp.Host("lms.domain.com")
        lms.GET("/dashboard", HomeLMS) // This will render the dashboard for the LMS

	return bapp
}

...
```

When testing, you need to specify the exact path, including the host, otherwise it will return a 404:

```
func (as ActionSuite) Test_App_Dashboard() {
    // This one will not throw a 404
    res := as.HTML("http://app.domain.com/dashboard/").Get()
    as.Equal(http.StatusOK, res.Code)
    as.Contains(res.Body.String(), "OK APP")
}

func (as ActionSuite) Test_LMS_Dashboard() {
    // This one will throw a 404
    res := as.HTML("http://cms.domain.com/dashboard/").Get()
    as.Equal(http.StatusOK, res.Code)
    as.Contains(res.Body.String(), "OK LMS")
}
```

Issue #1653, opened 2019-04-19
Application singleton is not really a singleton.


twisted1919:
## Description

It seems that the application singleton is not really a singleton, or something weird happens with my app.   

If you take a look at:
<img width="1123" alt="Screenshot 2019-04-19 20 18 21" src="https://user-images.githubusercontent.com/957490/56435234-6fd5f500-62e0-11e9-8798-49227c9bafda.png">
you will see how some of my routes are registered twice.  
I have a hunch this happens because in my app, i am also running a udp server, and inside it i defined a logger function like: 
```
func logger() buffalo.Logger {
	return actions.App().Options.Logger
}
``` 
And i guess this somehow affects what's going on. It's just a hunch though.  

However, if instead of:
```
    func App() *buffalo.App {
        if app == nil { ... }
    }
```

i'm using `sync.Once`:
```
    var appOnce sync.Once
    func App() *buffalo.App {
        appOnce.Do(func() {  ...  })
    }
```

Then my issue goes away, no more duplicate routes.  

### Steps to Reproduce the Problem

  See above.

### Expected Behavior

I expect to see unique routes.

### Actual Behavior

I see duplicate routes

sio4:
Indeed, this issue (`action.App()` is called multiple times while `app` is still `nil`) could very rarely happen since we don't use any locking method such as `sync.Once` on our scaffold. Also, `App` itself is not a singleton from the perspective of the core library, but it could be or have to be from the perspective of the user application.

However, it will not happen frequently (actually not happen in most case) since our scaffold also has `init()` function calling `actions.App()` in `grifts/init.go` before `main()` is executed. (this example [1] too) In this case, the variable `actions.app` will not be `nil` when `actions.App()` is called from other places including the `main()` of the app, but yes, it is not guaranteed when the app is somewhat complex.

But I am not sure how it happened... if the user-defined `logger()` which calls `App()` is called... from where and when...

I think we can improve the scaffolded `app.go` with `Sync` but still, this is a user's domain.

[1] http://gobuffalo.io/documentation/guides/goth/#example-usage

fasmat:
> Indeed, this issue (`action.App()` is called multiple times while `app` is still `nil`) could very rarely happen since we don't use any locking method such as `sync.Once` on our scaffold. Also, `App` itself is not a singleton from the perspective of the core library, but it could be or have to be from the perspective of the user application.
> 
[...]

PR #2239, opened 2022-04-16
v0.18.6

paganotoni:


app.go
```diff
@@ -14,15 +14,13 @@ import (
...
-	Middleware    *MiddlewareStack `json:"-"`
-	ErrorHandlers ErrorHandlers    `json:"-"`
-	router        *mux.Router
-	moot          *sync.RWMutex
-	routes        RouteList
-	root          *App
-	children      []*App
-	filepaths     []string
+	// Middleware, ErrorHandlers, router, and filepaths are moved to Home.
+	Home
```
fasmat, line 18: This is a breaking change, should we add fixers to update existing code?

grifts.go
```diff
@@ -44,33 +44,35 @@ func printMiddleware(a *App) {
 
 func printMiddlewareByRoute(a *App) {
 	mws := map[string]string{}
+	// TODO: middleware is 'per App' can it be a loop for Apps?
```
fasmat, line 47: I thought App is supposed to be a singleton? There is even an open issue that in some cases it is not when it should be: https://github.com/gobuffalo/buffalo/issues/1653

Now there can be multiple instances of app within the same binary? Isn't this a major breaking change?

route_mappings.go
```diff
@@ -254,22 +240,54 @@ func (a *App) ANY(p string, h Handler) {
...
 	g.RouteNamer = a.RouteNamer
 	g.Middleware = a.Middleware.clone()
 	g.ErrorHandlers = a.ErrorHandlers
-	g.root = a
-	if a.root != nil {
-		g.root = a.root
-	}
+
+	g.app = a.app  // will replace g.root
+	g.root = g.app // will be deprecated
```
fasmat, line 260: since g.root is a private field it could already be deprecated, while the public fields of `App` should not be removed but only marked as deprecated.

fasmat (changes requested):

app.go
sio4: could you please provide some more details about what is broken by this?
Actually, this patch (#2226) fixed the breaking change introduced by #2185. (more details in #2222)

fasmat: Any code that accesses `App.Middleware` or `App.Errorhandlers` will now fail to compile when upgrading from `v0.18.5` to `v0.18.6`

grifts.go
sio4: The `App` is not a singleton even though the name could be interpreted as such. Previously, (also still) `App.Group()` and `App.Resource()` return a new `App` which is a child of the root App. So as the result, there are many apps under the main App. The problem is that any instance of `App` still has the whole function of `App` so it is confusing and users can write a code like `App.Group().Stop()` which stops the root app.

This PR (#2226) did not change much and kept the structure the same as today since I didn't want to break any existing apps using v0, but it could be changed in the future version v1.

This PR is an interim solution for moving the app structure from `App` has `App`s to `App` has `Home`s. (I think the name `Home` is not super perfect and I thought `Group` could be the best, but we already have `App.Group` as a function name)

See #2222 and #2226 for more details and context.

route_mappings.go
sio4: As far as I remember, there is no deprecated *public* field. As you said, the private field could be deprecated right now but just keep them since (1) they are used in some other places and (2) want to track the change for v1 in the code context.

This change was introduced to fix the breaking change introduced by #2185 as a quick fix (with the smallest changes) and keep buffalo's existing behavior the same as now, and I think more effect could be focused on v1 with unavoidable breaking structure change.

app.go
sio4: Oh, really, I will check. I think they are still accessible since they are moved to `Home` which is embedded by `App`.

sio4: With my testing, it still works fine and I think there is no reason accessing `App.Middleware` is broken. See [1]

However, it seems like my buffalo binary seems like broken. `buffalo routes` and `buffalo task middleware` are not working with error:

```
.grifter/main.go:9:2: package grifts is not in GOROOT (/opt/google/go/src/grifts)
```

[1] https://github.com/gobuffalo/buffalo/pull/2239/files#diff-5b9f0b55bb5e0ae16b6fde53332a8923b6d445a022da239a5d23e66e7c7082e5R35

sio4: Works fine as below. The issue with `buffalo routes` and `buffalo task` seems related to `buffalo new app --skip-pop --skip-webpack --api`. With the basic webapp, no issue with them. (Will look at that issue later. It seems like github is not working properly now)

```console
$ buffalo tasks middleware
-> /
	github.com/gobuffalo/mw-csrf.glob..func1
	github.com/gobuffalo/buffalo-pop/v3/pop/popmw.Transaction.func2
	github.com/gobuffalo/mw-i18n/v2.*Translator.Middleware.func1
$ vi go.mod
$ go mod tidy
$ buffalo tasks middleware
-> http://127.0.0.1:3000/
	github.com/gobuffalo/mw-csrf.glob..func1
	github.com/gobuffalo/buffalo-pop/v3/pop/popmw.Transaction.func2
	github.com/gobuffalo/mw-i18n/v2.*Translator.Middleware.func1
$ git diff actions
diff --git a/actions/app.go b/actions/app.go
index be86b4c..3552ab2 100644
--- a/actions/app.go
+++ b/actions/app.go
@@ -52,6 +52,8 @@ func App() *buffalo.App {
                // Log request parameters (filters apply).
                app.Use(paramlogger.ParameterLogger)
 
+               app.Middleware.Clear()
+
                // Protect against CSRF attacks. https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)
                // Remove to disable this.
                app.Use(csrf.New)
$ 
```

fasmat: I assumed embedding home would cause a similar issue like the original problem, but thanks for double checking.

fasmat (approved):

PR #2252, opened 2022-04-28
Remove ioutil package

fasmat:
`ioutil` is deprecated for a few Go versions now. We should use the equivalent `io` and `os` functions instead.


sio4 (approved):
:+1:

PR #2260, opened 2022-05-05
Reduce dependencies on external libraries

fasmat:
This removes dependencies to the following externals:

* `github.com/karrick/godirwalk` - the standard library covers this functionality sufficiently well
* `github.com/markbates/oncer` - functionality is trivial, just used `sync.Once` instead
* `github.com/markbates/safe` - Inlined in one place, removed in another (code doesn't panic so no need to "safe" it)
* Various smaller issues that `go-golangci-lint` complains about


fasmat:
@paganotoni: ready for review :)

options.go
```diff
@@ -29,9 +28,7 @@ type Options struct {
 	// Env is the "environment" in which the App is running. Default is "development".
 	Env string `json:"env"`
 
-	// LogLevel defaults to "debug". Deprecated use LogLvl instead
-	LogLevel string `json:"log_level"`
```
sio4, line 33: deprecated 4 years ago and now dropped! congratulations, but I think the release note should have a comment on this dropping.

sio4 (approved):
LGTM, but it includes (minor) breaking change, so in my opinion, we need to be careful when we write release notes for the next release with it.

PR #2259, opened 2022-05-05
fix template based i18n support issues

sio4:
During upgrading my old app, I found some i18n features related to the templates were not working correctly. The issues are:

- When the `r.HTML(string)` stays with the old format, such as `r.HTML("index.html")`, but the template files are forms of `index.plush.html` and `index.plush.ko-kr.html` (renamed by `buffalo fix`), always the last file is used even though the browser's accept-language has no `ko` or `ko-KR`.
  - It happens because the logic of `templateRenderer.updateAliases()` does something unintended.
- Also, it is not a functional issue but `templateRenderer.resolve()` is called twice in many cases since `.localizedName()` uses this for checking if there is a template for possible languages.

#### Test environment

- template files:
  - `templates/index.plush.html`
  - `templates/index.plush.ko-kr.html`

**Debugging result when using `return c.Render(http.StatusOK, r.HTML("index.plush.html"))`**
--> OK, the correct templates are determined by the *direct name match*.

<details>

For ko-KR:

```
1. languages: [ko-KR ko en en-US en-US]
3. `index.plush.ko-kr.html`: try to resolve...
4. `index.plush.ko-kr.html` was **resolved directly** by the name
2. `index.plush.html` was localized as `index.plush.ko-kr.html`
3. `index.plush.ko-kr.html`: try to resolve...
4. `index.plush.ko-kr.html` was **resolved directly** by the name
3. `application.ko-kr.html`: try to resolve...
3. `application.ko.html`: try to resolve...
3. `application.en.html`: try to resolve...
2. `application.html` was localized as `application.html`
3. `application.html`: try to resolve...
5. `application.plush.html` was *resolved by map* for `application.html`
3. `_flash.html`: try to resolve...
5. `_flash.plush.html` was *resolved by map* for `_flash.html`
```

For the default language:
```
1. languages: [en en-US]
3. `index.plush.en.html`: try to resolve...
2. `index.plush.html` was localized as `index.plush.html`
3. `index.plush.html`: try to resolve...
4. `index.plush.html` was **resolved directly** by the name
3. `application.en.html`: try to resolve...
2. `application.html` was localized as `application.html`
3. `application.html`: try to resolve...
5. `application.plush.html` was *resolved by map* for `application.html`
3. `_flash.html`: try to resolve...
5. `_flash.plush.html` was *resolved by map* for `_flash.html`
```

</details>

**Debugging result when using `return c.Render(http.StatusOK, r.HTML("index.html"))`**
[...]

PR #2262, opened 2022-05-07
using path rather than filepath for URL-ish internal expressions

sio4:
Internal expressions of paths should use `path` instead of `path/filepath`.
- The main issue was on `htmlAutoRenderer.Render()` which calls like `ir.HTML("dir\file")` in Windows.
- Added some more checks on testcases for `buffalo.FS`'s `Open()`.
- Functions `assetPathFor()` and `loadManifest()` are also fixed to make their jobs clear.
- Additionally, fixed returning error for `ErrNotExist` state of `Open()`. 

It should be fixed before #2259.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -26,13 +26,28 @@ func NewFS(embed fs.ReadDirFS, dir string) FS {
 	}
 }
 
-// Open implements the FS interface.
+// Open opens the named file.
+//
+// When Open returns an error, it should be of type *PathError with the Op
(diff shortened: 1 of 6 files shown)


sio4:
Hi @paganotoni, could you please take a look at this PR? it is required to fix #2259.

paganotoni (approved):
LGTM. Thanks @sio4

sio4:
Thanks!

PR #2271, opened 2022-05-21
replace the package `refresh` with the forked gobuffalo version

sio4:
This PR will replace the `refresh` package with the forked gobuffalo version of `refresh`.


paganotoni (approved):
LGTM

Issue #2242, opened 2022-04-18
refine/improve the design of worker and simple worker implementation


sio4:
The current design and implementation of `Worker` and `Simple` worker are somewhat ambiguous and they need to be redefined. The points of consideration could be:

* Definition of Worker and its components including functions
* How to make them more reliable (when starting, stopping, and while running)
* Workflow (both automatic parts and manual parts)

#### TODO
* How to check if the worker is ready to work
  * When the worker should be initialized?
  * How to check if the worker is ready?
  * and purpose of the context
* Define the purpose of `Start(context)` clearly
* Consider timeout and/or cancellation for background jobs
* Consider if the already-submitted jobs to be performed even if the worker canceled before they are executed

#### Context
* #2241 (but not directly related)
* #2198

acaloiaro:
@sio4 I've been working on something that might be relevant to this issue: https://github.com/acaloiaro/neoq

It's still immature, but progressing nicely. With its current `MemBackend` and `PgBackend` implementations, I can see it being a reasonable replacement for `SimpleWorker` and `gocraft/work`. More out-of-the-box backends will be in the works for data stores like Redis, SQS, RabbitmQ, etc.

PR #2308, opened 2022-09-05
Extend `worker.Worker` to support periodic job registration

acaloiaro:
Extend `worker.Worker` to support periodic job registration, in support of https://github.com/gobuffalo/gocraft-work-adapter/issues/6

The purpose of this PR is to invite commentary on adding periodic job support. A concrete `worker.Simple` implementation would follow, along with tests.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -24,6 +24,8 @@ type Worker interface {
 	PerformIn(Job, time.Duration) error
 	// Register a Handler
 	Register(string, Handler) error
+	// RegisterPeriodic performs a job periodically according to the provided cron spec
+	RegisterPeriodic(cronSpec, jobName string, h Handler) error
 }
 
 /* TODO(sio4): #road-to-v1 - redefine Worker interface clearer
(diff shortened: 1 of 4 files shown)


sio4:
Please change the merge target to `v1`. The main branch is not yet ready.

sio4:
Hi @acaloiaro, sorry for my late response.

It seems like you worked based on the `main` branch, which is not yet prepared. (I am working on setting the next version's tree on there, but not yet finished) Please rebase your fix on the `v1` branch. Also, it could be better if you run the basic test on your local env before pushing the change.

acaloiaro:
@sio4 

I went ahead and rebased, and provided a concrete `Simple` implementation using `github.com/robfig/cron` as the cron implementation. This is the same dependency underlying https://github.com/gocraft/work and consequentially https://github.com/gobuffalo/gocraft-work-adapter

worker/worker.go
```diff
@@ -24,6 +24,8 @@ type Worker interface {
 	PerformIn(Job, time.Duration) error
 	// Register a Handler
 	Register(string, Handler) error
+	// RegisterPeriodic performs a job periodically according to the provided cron spec
+	RegisterPeriodic(cronSpec, jobName string, h Handler) error
```
sio4, line 28: First, this is a breaking change invalidates all existing (official/unofficial/custom) implementations. It should be considered as a separate interface or extended when we redesign the Worker support in a future major version.
(will add more comments in #2351)

acaloiaro:
Closing for now while a more holistic design is considered in #2242

PR #2327, opened 2022-09-18
Feature/defaults parameter

mohammadjafarsafarali:
Defaults functions support 2 items as parameter, in this way can handle more items

### DIFF
--- a/internal/defaults/defaults.go
+++ b/internal/defaults/defaults.go
@@ -12,11 +12,16 @@ func String(s ...string) string {
 	return r
 }
 
-func Int(i1, i2 int) int {
-	if i1 == 0 {
-		return i2
+func Int(i ...int) int {
+	var r int
+	for _, i := range i {
+		if i != 0 {
+			r = i
+			break
+		}
 	}
-	return i1
+
+	return r
 }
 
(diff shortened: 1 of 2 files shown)


sio4:
Hi @mohammadjafarsafarali, thank you for your contribution.

I like this extended method, but I wonder what is or could be the purpose of this change. These are `internal` functions so they will not be able to use outside. What is the background context of this PR?

mohammadjafarsafarali:
Thank @sio4, for taking the time to review my PR, 
actually I think we can have extended functions for later release and future changes.
In my opinion defaults package is somthing like helper functions and its completeness can help us for laters

In the end, your decision is respected

sio4:
I quickly checked the code base and found the functions are used in the following files:

* errors.go
* internal/httpx/content_type.go
* options.go
* method_override.go

and none of them used it more than once to get the same value which means there is no requirement of considering the value from multiple sources. (it looks like "A or B" is enough for now)

since I am freezing Buffalo core library, this is not good timing to improve those functions. Also, I think the direction of `defaults` could be something different in the future if we start considering a `generic` approach.

Thank you for this proposal, but I am going to close this.

PR #2328, opened 2022-09-18
fix: app.go LoadPlugins() function needs have error handling

mohammadjafarsafarali:
LoadPlugins() function in app.go file, is a must. this PR adds some error handler for this step

### DIFF
--- a/app.go
+++ b/app.go
@@ -35,7 +35,11 @@ func (a *App) Muxer() *mux.Router {
 
 // New returns a new instance of App and adds some sane, and useful, defaults.
 func New(opts Options) *App {
-	LoadPlugins()
+	err := LoadPlugins()
+	if err != nil {
+		panic("plugins load error")
+	}
+
 	envy.Load()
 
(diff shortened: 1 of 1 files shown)


sio4:
Which situation does it require to panic?

mohammadjafarsafarali:
Actually whenever LoadPlugins() function cant do the job, the app should panic because this process is necessary

sio4:
Sorry for my late response. I was busy with the other modules.

> Actually whenever LoadPlugins() function cant do the job, the app should panic because this process is necessary

It depends on each application's design. It could be better to check the result of the function call, and also better to allow the developer can handle their own tasks based on the result, but this is not a thing that should panic.

Also, even though the plugin interface exists, there is no official plugin that supports this event feature.

I wonder which is the use case you have, and what kind of issue occurred during the plugin loading. I need to know what is the real situation before assessing whether this PR could be ok.

Please provide more detailed information about the use case and the failing situation.

Issue #2335, opened 2022-09-29
document it! order of the default middleware and the logging behavior


sio4:
See #2300 and #2334

The behavior of default middleware (especially request logger) should be documented in detail.

Issue #2300, opened 2022-07-30
HTTP response code, `err` returned by request handlers, and middlewares


sio4:
From the definition of buffalo `Handler`: https://pkg.go.dev/github.com/gobuffalo/buffalo#Handler

> It is the responsibility of the Handler to handle the request/response correctly.

This statement also includes the responsibility of setting the HTTP response status correctly. However, sometimes we could omit to set the response code and just returns with `nil` or non-`nil` `err` which indicates the request was successfully completed (could be 200 by default) or failed (could be 500 by default). In this case, the `buffalo.Response` will not be configured completely, and it could make some confusion on middlewares and the result of what middlewares do.

#### One effect of this issue is https://github.com/gobuffalo/buffalo-pop/issues/19

In this case, the handler did its database job correctly but just returned `nil` without the setting response. As a result, even though the response to the client became 200 (since the returned error is `nil`) but the `pop` middleware detected it as a failure and the database rollback happened.

#### Another issue affected by this behavior is https://gophers.slack.com/archives/C3MSAFD40/p1658983223965079
(Coping from the original description):

> Hello! Question regarding something I've noticed when logging via the app error handlers, e.g. `app.ErrorHandlers[http.StatusInternalServerError] = someFunc`. I noticed that the log `status` gets set to 0 if the response has has not yet been written with the relevant status code prior to execution of the Error Handler callback. Here is a log item:
> ```
> time="2022-07-28T04:32:02Z" level=error msg="An internal error occurred: unhealthy" content_type=application/json duration="17.711µs" human_size="0 B" method=GET params="{}" path=/v2/health/ request_id=86619d31654c9c78b952-06d343033a282f7ef78e size=0 status=0
> ```
> The error handling itself is behaving as I would expect, and it is executing my custom error function as desired. I am then writing the appropriate status code to the response and the client sees it as a 500 error. However, because the log is setting a status of 0, my log aggregators pick this up with `EMERGENCY` level logging... which is not desirable. I verified that writing the response error to the response body before the error gets handled by `app.ErrorHandlers` sets the status code of the log appropriately. In the above example, it would be `status=500`.

(for the reference) My answer to this issue was:

> <...> Let's start with the output of `buffalo task middleware` from my test project:
> ```
> $ buffalo task middleware
> -> http://127.0.0.1:3000/
> 	github.com/gobuffalo/buffalo.*App.defaultErrorMiddleware
> 	github.com/gobuffalo/buffalo.*App.PanicHandler
[...]

PR #2345, opened 2022-10-25
added assert middleware to assert handler's behavior. (fix #2339)

### LINKED ISSUE #2339: meta issue for empty response status issue
#### Summary

From the definition of buffalo Handler: https://pkg.go.dev/github.com/gobuffalo/buffalo#Handler

> It is the responsibility of the Handler to handle the request/response correctly.

However, the responsibility of setting the status code on a response is slightly tricky, since the behavior of the standard [ResponseWriter](https://pkg.go.dev/net/http#ResponseWriter) allows developers can omit the status code when it is `200`. Even though we can have our own policy, aligning with the standard library could be the best for developers so they can expect the same behavior.

This issue is not just about responsibility but it affects the consistency of request processing. When if the developer omits the status code on their response,

* The default error middleware doesn't care about the status code if the returned error is nil
* The default request logger middleware just logs the status code as is, so it will be `0` but the client will get `200`
* The Pop middleware will rollback database transactions when the status code is not set as `2xx` or `3xx`

I think the direction should be aligned with the standard library, so we need to fix:

* make the default request logger logs the status code as `200` when the value is still `0`.
   * logging the details about auto set could be better but it is optional
* make the Pop middleware treats zero status code as success (actually informational 1xx should be the same)

#### Related issues and PRs

* [ ] #2300 --> incorrect information
   * [ ] #2334
* [ ] gobuffalo/buffalo-pop#19 --> inconsistent/contradictory response status and database action
   * [ ] gobuffalo/buffalo-pop#23
* [ ] gobuffalo/buffalo-pop#25 --> incorrect information

#### Related discussions:
* https://gophers.slack.com/archives/C3MSAFD40/p1658983223965079

@gobuffalo/core-managers what do you think? any other opinions?
This is a behavior change, but I don't think it will break existing apps widely. I will work on it if you have no concerns.

sio4:
For the [net/http#ResponseWriter](https://pkg.go.dev/net/http#ResponseWriter), `Write()` will set the response status as "200 OK" if a handler exited without calling `WriteHeader()` but returns `nil`. However, the Buffalo's behavior is slightly different. The Buffalo's actual responding behavior is basically the same since it uses the same, but the buffalo's middlewares (especially `poptx` from `buffalo-pop`) could treat them as a failed action when no response status is found. This inconsistency made some issues and this PR will fix the issue.

See more details in #2339 and other linked issues.

fixes #2339
fixes gobuffalo/buffalo-pop#25
fixes gobuffalo/buffalo-pop#19

related issues/PRs: #2300, #2334, #2335,

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -1,6 +1,7 @@
 package buffalo
 
 import (
+	"net/http"
 	"reflect"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


paganotoni (changes requested):
I like it. @sio4 Is there a chance we could add some tests to leave this behavior specified and guaranteed after v1.1.0 ?

sio4:
> I like it. @sio4 Is there a chance we could add some tests to leave this behavior specified and guaranteed after v1.1.0 ?

Sure, I agree. Will do that soon. (Maybe by this week)

paganotoni (approved):
LGTM!

PR #2346, opened 2022-10-25
removed pop logger setting that is not required to do by buffalo

sio4:
We don't need to configure a pop logger here. By default, the logger used for the app generated by buffalo will use the following logger if they are configured with pop middleware. This is redundant and it made the buffalo core depends on pop.

https://github.com/gobuffalo/buffalo-pop/blob/e56c88988c60c8567cb7da927354c7d734665b20/pop/popmw/tx.go#L29

Now, we completely decoupled buffalo and pop.


paganotoni (approved):
This is great @sio4. Thanks for it.

Issue #1904, opened 2020-02-13
Error Trace Doesn't Include Stack Trace


stnguyen90:
### Description

The documentation on [Error Handling](https://gobuffalo.io/en/docs/errors/) indicates there should be a stack trace, but stack traces are no longer returned.

### Steps to Reproduce the Problem

1. `buffalo new coke`
1. `cd coke`
1. `buffalo dev &`
1. `curl -H 'Content-Type: application/json' 'http://localhost:3000/does_not_exist'`

### Expected Behavior

`trace` and `error` should not be the same. `trace` should have the stack trace.

### Actual Behavior

{"error":"could not find does_not_exist","trace":"could not find does_not_exist","code":404}

### Info

Please run `buffalo info` and paste the information below where it says "PASTE_HERE".

<details>

```
-> Go: Checking installation
✓ The `go` executable was found on your system at: /usr/local/go/bin/go

-> Go: Checking minimum version requirements
✓ Your version of Go, 1.13.7, meets the minimum requirements.

-> Go: Checking Package Management
✓ You are using Go Modules (`go`) for package management.

-> Go: Checking PATH
✓ Your PATH contains /go/bin.

-> Node: Checking installation
✓ The `node` executable was found on your system at: /usr/bin/node

-> Node: Checking minimum version requirements
✓ Your version of Node, v10.19.0, meets the minimum requirements.

-> NPM: Checking installation
✓ The `npm` executable was found on your system at: /usr/bin/npm

-> NPM: Checking minimum version requirements
✓ Your version of NPM, 6.13.4, meets the minimum requirements.

-> Yarn: Checking installation
✓ The `yarnpkg` executable was found on your system at: /usr/bin/yarnpkg

-> Yarn: Checking minimum version requirements
✓ Your version of Yarn, 1.22.0, meets the minimum requirements.

-> PostgreSQL: Checking installation
✘ The `postgres` executable could not be found on your system.
For help setting up your Postgres environment please follow the instructions for you platform at:

https://www.postgresql.org/download/

-> MySQL: Checking installation
✘ The `mysql` executable could not be found on your system.
For help setting up your MySQL environment please follow the instructions for you platform at:

https://www.mysql.com/downloads/

-> SQLite3: Checking installation
✓ The `sqlite3` executable was found on your system at: /usr/bin/sqlite3

-> SQLite3: Checking minimum version requirements
✓ Your version of SQLite3, 3.27.2, meets the minimum requirements.

[...]

PR #2358, opened 2022-12-30
Piti/feat/trace on dev test

piti118:
### What is being done in this PR?

This PR fixes issue #1904. And fix the bug caused by one of the previous pr that trace will only appear on development but not test environment.

### What are the main choices made to get to this solution?

Without traces it's much harder to debug. The doc is also misleading that we will get a trace but in fact we didn't. This PR fix that issue.

https://gobuffalo.io/documentation/request_handling/errors/#default-error-handling-development

The tests are covered in unit tests.

On dev, test: when handler return error being (pkg/errors or context.Error()) if it contains stack trace then it should return that stack trace.
On other env, trace is hidden.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -15,6 +15,7 @@ import (
 	"github.com/gobuffalo/buffalo/internal/httpx"
 	"github.com/gobuffalo/events"
 	"github.com/gobuffalo/plush/v4"
+	pkgError "github.com/pkg/errors"
 )
 
 // HTTPError a typed error returned by http Handlers and used for choosing error handlers
@@ -34,6 +35,16 @@ func (h HTTPError) Error() string {
 		return h.Cause.Error()
 	}
 	return "unknown cause"
+}
+
+func (h HTTPError) StackTrace() pkgError.StackTrace {
+	if h.Cause == nil {
+		return nil
+	}
+	if tracer, ok := h.Cause.(stackTracer); ok {
+		return tracer.StackTrace()
+	}
+	return nil
 }
 
 // ErrorHandler interface for handling an error for a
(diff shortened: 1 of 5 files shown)


errors.go
```diff
@@ -165,8 +176,27 @@ type ErrorResponse struct {
...
+	if errStack, ok := err.(stackTracer); ok {
+		return fmt.Sprintf("%+v", errStack.StackTrace())
+	} else {
+		return err.Error()
+	}
+}
+
+func isUnsafeEnvironment(c Context) bool {
 	env := c.Value("env")
+	return env != nil && env.(string) != "development" && env.(string) != "test"
```
sio4, line 195: the environment "test" should execute the same code set as "production" (with data for testing) and this is not acceptable.

sio4 (commented):
First, even though it's not a concrete rule, it could be better if we don't rely on specific third-party packages as much as possible and we previously removed usages of `pkg/errors` before. See old PRs to check the history of tracing features.

I mostly agree that having a stack trace is better for debugging the user's application but supporting it from the perspective of the framework could be made with a different approach without directly supporting a specific third party. I was working on the related code before but could not make it a high priority before. It would be better if it is flexible enough, I am also considering better support of events, and I will work on it soon.


Related PRs:
[1] #138
[2] #1643
[3] #1930
[4] #2352

sio4:
Hi @piti118,
Could you please check if PR #2361 covers your needs for stack trace? It will print out stack trace when if the user's handler returns an `error` that supports `%+v` like the error generated with `pkg/errors`.

sio4:
Hi, please check #2361 for the current status. I just merged the PR into the v1 branch and I am going to close this PR to proceed to the next release.

Issue #2356, opened 2022-12-26
bug: buffalo.Options.Prefix does not work


qdongxu:
#### Description

 Buffalo version is: v0.18.12

I want the buffalo server as a backend of nginx, and follow the instruction from #2022 , adding a Options.Prefix: '/prefix', then all request returns 404 no matter a '/prefix' is prepended in the request or not.  

I found this useful guideline to configure with the nginx https://gobuffalo.io/documentation/deploy/proxy/ , but the rule inside my organization requires a prefix route config. 

#### Expected Behavior

Expected to configure a reverse proxy route with prefix .

#### Actual Behavior
web requests are broken after the prefix is configured. 
API requests should append a '/' char at the end of the path.

### To Reproduce

1. Add a prefix in `actions/App.go`

```go
func App() *buffalo.App {
	appOnce.Do(func() {
		app = buffalo.New(buffalo.Options{
			Env:         ENV,
			SessionName: "_tmon_session",
			Prefix:      "/prefix/",
		})
}
```

2. run `buffalo routes` to verify:

before adding prefix:

```
  % buffalo routes
METHOD | HOST                  | PATH                    | ALIASES | NAME                   | HANDLER
------ | ----                  | ----                    | ------- | ----                   | -------
GET    | http://127.0.0.1:3000 | /                       |         | rootPath               | proj/actions.HomeHandler
POST   | http://127.0.0.1:3000 | /api/v1/user/users/     |         | apiV1UserUsersPath     | proj/actions/user.CreateUserHandler
```

after adding the prefix:

```
% buffalo routes
METHOD | HOST                  | PATH                           | ALIASES | NAME                         | HANDLER
------ | ----                  | ----                           | ------- | ----                         | -------
GET    | http://127.0.0.1:3000 | /prefix/                       |         | prefixPath                   | proj/actions.HomeHandler
POST   | http://127.0.0.1:3000 | /prefix/api/v1/user/users/     |         | prefixAPIV1UserUsersPath     | proj/actions/user.CreateUserHandler
```
4. check the welcome page, the returned links do not contain the '/prefix':
```
% curl -L  http://127.0.0.1:3000/prefix
<pre>
<a href="assets/">assets/</a>
<a href="robots.txt">robots.txt</a>
</pre>

# but http://127.0.0.1:3000/prefix/ , with an ending '/' , returns 404.
```

[...]

PR #2366, opened 2023-02-19
fix URL Prefix

qdongxu:
This PR addresses Issue #2356. 

The web console cannot be opened after setting a URL prefix in `actions/app.go` :
```go
	appOnce.Do(func() {
		app = buffalo.New(buffalo.Options{
			Env:         ENV,
			SessionName: "_prefix_test_session",
			Prefix:      "/prefix/",
		})

```
The error is : 
```
home/index.plush.html: line 5: "rootPath": unknown identifier
```

`routenamer.go` consider's "/" as the root path only, ignoring a URL prefix as a root path.

1. Add a prefix as above then open `http://127.0.0.1:3000/prefix/` in the browser.

### DIFF
--- a/app.go
+++ b/app.go
@@ -56,7 +56,9 @@ func New(opts Options) *App {
 		routes:   RouteList{},
 		children: []*App{},
 
-		RouteNamer: baseRouteNamer{},
+		RouteNamer: baseRouteNamer{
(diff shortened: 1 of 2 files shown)


qdongxu:
closed as submitted to the wrong branch.

Issue #2240, opened 2022-04-17
make the `App` the root of whole application and introduce `Home` as a subapp


sio4:
The name `App` sounds like the only root of the application structure (could be a singleton) but actually, it is also a container of a group of URLs that is independently configurable with middlewares and error handlers. The `App.Group()` and `App.Resource()` return a new `App` which is a child of the root `App`, it's OK, but they have the whole functions of `App` which is not desirable.

By #2226, a quick fix for #2185, I changed some of the internal implementations of `App` while keeping the external behavior the same, but there are still some issues caused by the current design. Since we are now discussing v1, the future of buffalo, so I would like to suggest that we keep this structure for the current version and improve the core when we move to the next version.

#### TODO

* make the `Home` the container of a group of URLs and its independent middlewares and error handlers (half done)
* make `Home` as a child of the `App`.
* make `App` as a singleton (but we need more consideration for this since it is not always a good thing.)

#### Context

* #2239
* #2226
* #2222
* #2185
* #1653

PR #2367, opened 2023-02-19
Fix URL prefix

qdongxu:
This PR addresses Issue #2356. 

The web console cannot be opened after setting a URL prefix in `actions/app.go` :
```go
	appOnce.Do(func() {
		app = buffalo.New(buffalo.Options{
			Env:         ENV,
			SessionName: "_prefix_test_session",
			Prefix:      "/prefix/",
		})

```
The error is : 
```
home/index.plush.html: line 5: "rootPath": unknown identifier
```

`routenamer.go` considers "/" as the root path only, ignoring a URL prefix as a root path.

1. Add a prefix as above then open `http://127.0.0.1:3000/prefix/` in the browser.

### DIFF
--- a/app.go
+++ b/app.go
@@ -56,7 +56,9 @@ func New(opts Options) *App {
 		routes:   RouteList{},
 		children: []*App{},
 
-		RouteNamer: baseRouteNamer{},
+		RouteNamer: baseRouteNamer{
+			Prefix: opts.Prefix,
+		},
 	}
 	a.Home.app = a     // replace root.
 	a.Home.appSelf = a // temporary, reverse reference to the group app.
--- a/render/options.go
+++ b/render/options.go
@@ -32,4 +32,7 @@ type Options struct {
 	// DefaultContentType instructs the engine what it should fall back to if
 	// the "content-type" is unknown
 	DefaultContentType string
+
+	// Prefix inherits the global single prefix from buffalo.Options
+	Prefix string
 }
--- a/render/template.go
+++ b/render/template.go
@@ -266,18 +266,18 @@ func (s *templateRenderer) assetPath(file string) (string, error) {
 		if err != nil {
 			manifest, err = s.AssetsFS.Open("assets/manifest.json")
 			if err != nil {
-				return assetPathFor(file), nil
+				return assetPathFor(s.Prefix, file), nil
 			}
 		}
 		defer manifest.Close()
 
 		err = loadManifest(manifest)
 		if err != nil {
-			return assetPathFor(file), fmt.Errorf("your manifest.json is not correct: %s", err)
+			return assetPathFor(s.Prefix, file), fmt.Errorf("your manifest.json is not correct: %s", err)
 		}
 	}
 
-	return assetPathFor(file), nil
+	return assetPathFor(s.Prefix, file), nil
 }
 
 // Template renders the named files using the specified
--- a/render/template_helpers.go
+++ b/render/template_helpers.go
@@ -45,12 +45,12 @@ func (s *templateRenderer) addAssetsHelpers(helpers Helpers) Helpers {
 
 var assetMap = stringMap{}
 
-func assetPathFor(file string) string {
+func assetPathFor(prefix string, file string) string {
 	filePath, ok := assetMap.Load(file)
 	if filePath == "" || !ok {
 		filePath = file
 	}
-	return path.Join("/assets", filePath)
+	return path.Join(prefix, "/assets", filePath)
 }
 
 func loadManifest(manifest io.Reader) error {
--- a/routenamer.go
+++ b/routenamer.go
@@ -17,9 +17,15 @@ type RouteNamer interface {
 }
 
 // BaseRouteNamer is the default route namer used by apps.
-type baseRouteNamer struct{}
+type baseRouteNamer struct {
+	Prefix string
+}
 
 func (drn baseRouteNamer) NameRoute(p string) string {
+	if drn.Prefix != "" {
+		p = strings.TrimPrefix(p, drn.Prefix)
+	}
+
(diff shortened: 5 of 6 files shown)


sio4:
This PR is not acceptable:

1. This changes the behavior and will cause unwanted results if existing users use the current result of names. (e.g. `prefixSomethingPath` instead `somethingPath` for the prefixed `/something`) This is a breaking change.
2. The fix just fixes the root path only. If the goal of this PR is to keep the route name the same regardless of the prefix, the fix should address the other paths too. (one possible way is trimming the prefix from the given path before the evaluation)

Basically, I agree that the path name should be preserved when we use the prefix to make the app can be ported (mounted) in a different prefix easily without code level change. However, we cannot change this behavior in v1. Also, I have the plan (#2240) to rewrite related things including App, Home, Group, and route information then this behavior could be redefined at that moment.

qdongxu:
> This PR is not acceptable:
> 
> 1. This changes the behavior and will cause unwanted results if existing users use the current result of names. (e.g. `prefixSomethingPath` instead `somethingPath` for the prefixed `/something`) This is a breaking change.
> 2. The fix just fixes the root path only. If the goal of this PR is to keep the route name the same regardless of the prefix, the fix should address the other paths too. (one possible way is trimming the prefix from the given path before the evaluation)
> 
> Basically, I agree that the path name should be preserved when we use the prefix to make the app can be ported (mounted) in a different prefix easily without code level change. However, we cannot change this behavior in v1. Also, I have the plan (#2240) to rewrite related things including App, Home, Group, and route information then this behavior could be redefined at that moment.

Hi @sio4, trimming the prefix from the route name, resolved both concerns.  This is a fully backward-compatible solution.

sio4:
> > This PR is not acceptable:
> > 
> > 1. This changes the behavior and will cause unwanted results if existing users use the current result of names. (e.g. `prefixSomethingPath` instead `somethingPath` for the prefixed `/something`) This is a breaking change.
> > 2. The fix just fixes the root path only. If the goal of this PR is to keep the route name the same regardless of the prefix, the fix should address the other paths too. (one possible way is trimming the prefix from the given path before the evaluation)
> > 
> > Basically, I agree that the path name should be preserved when we use the prefix to make the app can be ported (mounted) in a different prefix easily without code level change. However, we cannot change this behavior in v1. Also, I have the plan (#2240) to rewrite related things including App, Home, Group, and route information then this behavior could be redefined at that moment.
> 
> Hi @sio4, trimming the prefix from the route name, resolved both concerns. This is a fully backward-compatible solution.

While the current code generates a route name as `prefixPath` and `prefixHomePath` but the patched version will generate `rootPath` and `homePath`. What do you mean by fully backward-compatible?

qdongxu:
> While the current code generates a route name as prefixPath and prefixHomePath but the patched version will generate rootPath and homePath. What do you mean by fully backward-compatible?

No. the code has been updated as generating a route name as `rootPath` and `homePath` without the prefix.
**fully backward-compatible** just means both `rootPath` and `homePath` keep unchanged despite a prefix is configured or not.

Input with a prefix:
```go

app = buffalo.New(buffalo.Options{
	Env:         ENV,
	SessionName: "_prefix_test_session",
	Prefix:      "/prefix",
})

app.GET("/", HomeHandler)

app.GET("/about/", About)
app.GET("/prefix/about/", About)

h1 := app.VirtualHost("h1.com")
h1.GET("/func1", VirtualHost1About)

h2 := app.VirtualHost("h2.com")
h2.GET("/func1", VirtualHost2About)

g1 := app.Group("/group/")
g1.GET("func2", About)

subg1 := g1.Group("subg")
subg1.GET("func3", About)
```
Route names are identical as no prefix configured:

METHOD | PATH | NAME | HANDLER
-- | -- | -- | --
GET | /prefix/ | rootPath | prefix_test/actions.HomeHandler
GET | /prefix/about/ | aboutPath | prefix_test/actions.About
GET | /prefix/prefix/about/ | prefixAboutPath | prefix_test/actions.About
GET | /prefix/func1/ | func1Path | prefix_test/actions.VirtualHost1About
GET | /prefix/func1/ | func1Path | prefix_test/actions.VirtualHost2About
GET | /prefix/group/func2/ | groupFunc2Path | prefix_test/actions.About
GET | /prefix/group/subg/func3/ | groupSubgFunc3Path | prefix_test/actions.About

sio4:
> > While the current code generates a route name as prefixPath and prefixHomePath but the patched version will generate rootPath and homePath. What do you mean by fully backward-compatible?
> 
> No. the code has been updated as generating a route name as `rootPath` and `homePath` without the prefix. **fully backward-compatible** just means both `rootPath` and `homePath` keep unchanged despite a prefix is configured or not.

No, what I mean is that the result should be preserved as the same as today (which is `prefixPath` instead of `rootPath`) for the current major version. **If we release a new version with this change without bumping the major version, the existing application that already uses the current route name as `prefixPath` will be broken when they update the module.** This is breaking change even though the current behavior is not good.

Also, your PR is targeted to the `main` branch but a PR against the branch is not acceptable for now. We are planning a new major version on the branch, but not today.

I would like to keep this PR as the status of blocked and will revisit it once we started working on the new major version soon.

sio4:
Also, in the next major version, the App/Home/Group structure could be completely changed as I already linked the open issue #2240. So I am not fully sure if we can merge this change directly or just need to fix the issue in a different way.

qdongxu:
Currently `rootPath()`, and `javascriptTag()`  in the `render` package do not work with Prefix.  the web console cannot open at all when a prefix is configured. The only to configure a reverse proxy is by routing by port.

It will be great if this issue is scheduled in #2240.  thanks for your patient clarification, @sio4

qdongxu:
This PR closed as it will be addressed in #2240

PR #2400, opened 2025-05-08
Avoid memory allocation when calling Render

Mido-sys:
We can avoid the extra head by using ```unsafe.Pointer``` when writing back in ```io.Writer```

### DIFF
--- a/render/template.go
+++ b/render/template.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"unsafe"
 
 	"github.com/sirupsen/logrus"
 	"golang.org/x/text/language"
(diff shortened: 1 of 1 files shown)


Mido-sys:
@paganotoni,

Just a kind reminder. A slight improvement.

paganotoni:
@Mido-sys this is great. Would it be possible to write some sort of benchmark for this?

Mido-sys:
@paganotoni .

For sure:

```BenchmarkHTTPHandlerSafe-20    	    7390	    319451 ns/op	 1033094 B/op	       9 allocs/op```
```BenchmarkHTTPHandlerUnsafe-20    	    9504	    147947 ns/op	  516989 B/op	       8 allocs/op```

I performed the test on a webpage that's around 500kb
1. Unsafe HTTP handler is >2× faster than the safe one and avoids extra allocs. 

1. Safe path allocates ~1 MB per request, while Unsafe cuts that in half (~517 KB)

paganotoni (approved):

paganotoni:
Thanks @Mido-sys!

